PR #5, opened 2016-02-28
Eatery Data Persistence!

dantheli:
Eateries are now stored on the phone. After seven days they are retrieved from the server again. DiningStack changed so do `pod install` again.


lucasderraugh:
If you can just remove the print statements (you can `git commit --amend` that) then you're good to go.

PR #31, opened 2016-03-04
Daniel/watch polish

dantheli:
Also added bullets to menu items


lucasderraugh:
I can almost guarantee we'll be changing those colors haha. That gray is just to damn close to that green.

PR #35, opened 2016-03-06
Kevin/ft quick actions

kevingreer:


Classes/Eat Now/Detail View Controller/EateriesGridViewController.swift
```diff
@@ -105,9 +106,39 @@ class EateriesGridViewController: UIViewController, UICollectionViewDataSource,
...
+      detailViewController.displayedDate = NSDate()
+      detailViewController.selectedMeal = nil
+      detailViewController.delegate = self
+      
+      // Unwind back to this VC if it is not showing
+      if !(self.navigationController?.visibleViewController is EateriesGridViewController) {
+          self.navigationController?.popToRootViewControllerAnimated(false)
+      }
+      
+      self.navigationController?.pushViewController(detailViewController, animated: false)
```
lucasderraugh: Preferably only insert `self.` when inside a closure that requires its use.

Issue #54, opened 2016-03-15
Add back in Pull to Refresh


lucasderraugh:
Now that we've moved the searchbar into a sane mechanism of display (out of the collectionview), add in some cool refresh control.

PR #55, opened 2016-03-16
Fancy pull to refresh

dantheli:
Addresses #54

### DIFF
--- a/Classes/Eat Now/Detail View Controller/EateriesGridViewController.swift
+++ b/Classes/Eat Now/Detail View Controller/EateriesGridViewController.swift
@@ -8,6 +8,7 @@
 
 import UIKit
 import DiningStack
+import DGElasticPullToRefresh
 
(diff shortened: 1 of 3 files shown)


lucasderraugh:
@dantheli I've been going back and forth in my head about if I find this good or annoying to have PTR (or at least this particular one). On the one hand it looks cool, but the other hand, everytime I scroll to the top the nav bar looks like it bounces just a little which I find somewhat annoying. Perhaps a better feature would be to avoid PTR and just guarantee the user has the most updated times for Eateries. Since this info only needs to be pulled on the daily I'm not sure if PTR is necessary. I think a better feature would be having the times update live (every minute we update the time remaining on the cells for example). Leave this open for now, I'll ask what you think of it tonight. I know users almost expect a PTR for the feed, but when it can be automatic I'm less inclined to make it a 'feature'.

PR #56, opened 2016-03-16
Kevin/sharing closed eateries

kevingreer:
The bug was with the bar in the menuHeader. It was calculating a width by dividing by the number of events, but since the events dictionary was empty, some fields were being set to NaN or inf.

### DIFF
--- a/Eatery/Eat Now/MenuImages.swift
+++ b/Eatery/Eat Now/MenuImages.swift
@@ -171,10 +171,13 @@ class MenuImages: NSObject {
         eateryNameLabel.center = CGPointMake(eateryNameLabel.center.x, openIndicatorView.center.y)
         
         //create eatery time label
-        let openTime = "Open \(events[selectedMenu]!.startDateFormatted) to \(events[selectedMenu]!.endDateFormatted)"
+        var text = "Closed"
+        if let event = events[selectedMenu] {
+            text = "Open \(event.startDateFormatted) to \(event.endDateFormatted)"
+        }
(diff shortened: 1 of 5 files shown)


Eatery/Eat Now/MenuImages.swift
```diff
@@ -171,34 +171,38 @@ class MenuImages: NSObject {
...
         eventNameLabel.font = headerEventFont
         eventNameLabel.sizeToFit()
         eventNameLabel.center = CGPointMake(menuHeader.center.x, menuHeader.frame.height - eventNameLabel.frame.height)
         menuHeader.addSubview(eventNameLabel)
-        
+      
         //create event name underscore bar view
-        let bar = UIView(frame: CGRectMake(0,menuHeader.frame.height - 2, menuHeader.frame.width / CGFloat(events.count), 2))
+        let sections = CGFloat(events.count < 1 ? 1 : events.count)
+        let bar = UIView(frame: CGRectMake(0,menuHeader.frame.height - 2, menuHeader.frame.width / sections, 2))
```
kevingreer, line 201: This was the line causing the error

Podfile
```diff
@@ -1,7 +1,7 @@
 use_frameworks!
 
 def shared_pods
-  pod 'DiningStack', :git => 'https://github.com/cuappdev/DiningStack.git', :branch => 'lucas/cocoapods'
+  pod 'DiningStack', :git => 'https://github.com/cuappdev/DiningStack.git', :branch => 'kevin/un-hardcoding_menus'
```
kevingreer: This should be changed when my DiningStack pull request is okayed

lucasderraugh:
@kevingreer If you've tested it to make sure it doesn't break the existing app, then I'm fine with merging this in. Could you remove the autoresizing mask sets to false though since we were trying that around for testing purposes, but I don't think you resize any of the existing views anyway so it shouldn't be in there.

PR #57, opened 2016-03-18
Jesse/more info

ChenJesse:
issue 53

Right now pretty barebones implementation with a UITableView, API doesn't offer the website so it's just displayed as "N/A" atm, not sure what we should replace it with. Info is a bit redundant but theres nothing more significant to really fill it in with (About usually summarizes Location, Whereabouts, Campus)


say25:
simulator screen shot mar 18 2016 3 48 41 am
Couple of things:
1. The back arrow should probably just simply say "back" rather than the date.
2. I would also merge Location and Campus To maybe say "West Campus \n Alice Cook House"
3. The above image is probably closer to the ideal order of information.

ChenJesse:
Embedded MKMapView

PR #65, opened 2016-03-23
Default Guide to Current Time

### LINKED ISSUE #60: Look Ahead should default to meal for current time on Today date

dantheli:
Currently [0:00, 11:00) = breakfast, [11:00, 16:00) = lunch, [16:00, 0:00) = dinner. These constants can be changed at the top of LookAheadViewController.swift. Resolves #60.

### DIFF
--- a/Eatery/LookAheadViewController.swift
+++ b/Eatery/LookAheadViewController.swift
@@ -15,6 +15,8 @@ private let DayDateFormatter: NSDateFormatter = {
     dateFormatter.dateFormat = "EEE"
     return dateFormatter
 }()
+private let LunchStartHour = 11
+private let DinnerStartHour = 16
 
 class LookAheadViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, FilterEateriesViewDelegate, EateryHeaderCellDelegate, FilterDateViewDelegate, EateryMenuCellDelegate {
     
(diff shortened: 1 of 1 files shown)


Eatery/LookAheadViewController.swift
```diff
@@ -421,4 +424,17 @@ class LookAheadViewController: UIViewController, UITableViewDataSource, UITableV
...
+        let calendar = NSCalendar.currentCalendar()
+        let currentHour = calendar.component(.Hour, fromDate: NSDate())
+        if currentHour < LunchStartHour {
+            return 0 // Breakfast
+        } else if currentHour < DinnerStartHour {
+            return 1 // Lunch
+        } else {
+            return 2 // Dinner
+        }
+    }
```
lucasderraugh, line 433: Alternatively:

``` swift
func currentMealIndex() -> Int {
    let currentHour = NSCalendar.currentCalendar().component(.Hour, fromDate: NSDate())
    switch currentHour {
    case 0...9: return 0 // Breakfast
    case 10...15: return 1 // Lunch
    default: return 2 // Dinner
    }
}
```

Also, I don't want to have private let properties if I only use them in this function.

lucasderraugh:
Don't really care if you use the range switch statement, but let's not pollute class with properties if they're local to one function.

lucasderraugh:
Good to go

Issue #53, opened 2016-03-15
More Details Screen for an Eatery


say25:
Use other data we get from Dining API. Ability to find Eatery on Map, send email, make phone call, etc.

say25:
Maps aspect of this issue handled with #67. Postponing all other aspects for now.

PR #67, opened 2016-03-26
Jesse/toggle map

ChenJesse:
Partial solution to issue #53, brings up MKMapView of eatery location


lucasderraugh:
Ideally we would self contain this in it's own VC. I'm ok with this being merged but if you'd like to work on it more that would be good too. 2 main things that stick out to me:
1. Let's start using callbacks instead of delegates when applicable.
2. Make the map view be a self contained VC that can dismiss itself on the button tap. Otherwise the MenuVC takes on that responsibility which is really not the place for it.

ChenJesse:
Made the changes you asked for.

lucasderraugh:
Merged in 2694e81

lucasderraugh:
@ChenJesse Just might want to note some of the dependency injection added to the class. The inits are ugly within the class, but because the class requires that eatery, it's nicer to inject the eatery object directly. You can initialize all the other objects ahead of time guaranteeing their existence instead of implicitly unwrapping them.

PR #71, opened 2016-04-13
Background with message for share menu

dantheli:


lucasderraugh:
Merged in 3274476

lucasderraugh:
@dantheli Made small tweaks in 2fe1c5f9b1. You can use type inference with class methods (with colors), there are simpler animations methods that infer delay and options to be 0 and empty. Overall really awesome work though simplifying some of these calls.

PR #73, opened 2016-04-14
Logan/guide enhancements2

loganallen:
Guide cell clicking enhancements


lucasderraugh:
Merged in 03c01c5

lucasderraugh:
@loganallen  Few fixes in 8f99bbc57 (which I somehow made you the author for, so you're welcome I guess). Basically abstracted out the array filtering lines so you pass in 2 arrays by reference and it does the logic that were in the 3 different switch cases. Please check it out and let me know if you have questions.

PR #75, opened 2016-04-14
Jesse/smoother scrolling

ChenJesse:


lucasderraugh:
Going to close this because there were issues with this implementation (mostly because the images are never cached). Using 3rd party library for now, which is pretty dope. See 1d05eaf for that change.

Issue #69, opened 2016-04-06
Emoji Search


say25:

anniexcheng:
Merged in 7eeff9b371e6f5e41eca96a262410a062d8b636e

PR #78, opened 2016-04-17
Finished emoji search method (without JSON)

myo3:
@say25 and team will add more emojis on the fly

### DIFF
--- a/Eatery/Eat Now/Detail View Controller/EateriesGridViewController.swift
+++ b/Eatery/Eat Now/Detail View Controller/EateriesGridViewController.swift
@@ -180,7 +180,7 @@ class EateriesGridViewController: UIViewController, MenuButtonsDelegate {
                 var currentMenuFoodItemFound = false
                 if let activeEvent = eatery.activeEventForDate(NSDate()) {
                     for item in activeEvent.menu.values.flatten() {
-                        if item.name.rangeOfString(searchQuery, options: options) != nil && eatery.slug == "North-Star"{
+                        if item.name.rangeOfString(searchQuery, options: options) != nil {
                             if self.searchedMenuItemNames[eatery] == nil {
                                 self.searchedMenuItemNames[eatery] = [item.name]
                             } else {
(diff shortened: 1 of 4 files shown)


Eatery/Eat Now/Detail View Controller/EateriesGridViewController.swift
```diff
@@ -156,9 +156,22 @@ class EateriesGridViewController: UIViewController, MenuButtonsDelegate {
...
+        }
+        
+        return modifiedSearchBarText
+    }
+    
     func processEateries() {
         var desiredEateries: [Eatery] = []
-        let searchQuery = searchBar.text ?? ""
+        let searchBarText = findEmojis(searchBar.text ?? "") //replace unicode emoji w/ text name
+        let searchQuery = searchBarText
```
anniexcheng: You could just combine this line with the previous line, so let searchQuery = findEmojis(searchBar.text ?? "").

Eatery/Eat Now/Detail View Controller/EateriesGridViewController.swift
```diff
@@ -156,9 +156,22 @@ class EateriesGridViewController: UIViewController, MenuButtonsDelegate {
         loadData(false, completion: nil)
     }
     
+    func findEmojis(searchBarText: String) -> String{
+        let emojiDictionary: [String: String] = ["🍔": "burger", "🍟": "fries", "🍕": "pizza", "🐔": "chicken", "🌮": "taco", "🌯": "burrito", "🍳" : "egg", "🍚" : "rice", "🍝" :  "spaghetti", "💩": "nasties"]
```
anniexcheng: It's probably better to put the emoji conversions into a separate json file (similar to nicknames.json) so we can add more in the future.

lucasderraugh: It would be better to leave the emoji stuff as a Swift file, but keeping it in a separate Swift file would be ideal.

anniexcheng:
Close #69

PR #82, opened 2016-04-17
Add calls to analytics tracking functions

eappel:
Revived AnalyticsManager.swift.  Updated tracking functions and added new ones for new features.   

This will allow us to get a better idea of how people are using the app, but more importantly, it will give us sexy pie charts to put in keynote presentations like the one below:  
screen shot 2016-04-17 at 4 47 18 pm

### DIFF
--- a/Eatery/AppDelegate.swift
+++ b/Eatery/AppDelegate.swift
@@ -63,7 +63,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let uuid = NSUUID().UUIDString
         SEGAnalytics.sharedAnalytics().identify(uuid)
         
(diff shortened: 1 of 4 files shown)


say25:
Maybe we should add analytics for lookahead (guide) to figure out how many days people lookahead and for what meals.

eappel:
@say25 This PR adds code to track viewing the "look ahead" screen but doesn't give any more granularity.   I thought about going more in depth but I'm not sure how that would help inform any design decisions in the future.

eappel:
I'm curious to see what we can learn from these preliminary points and then we can always add more in the next update

lucasderraugh:
Merged in 3b30a8d

PR #84, opened 2016-04-21
Added to Favorites popup. Changed favorites star to solid gold

myo3:


say25:
Missing at commit to link all the files together

say25:
Maybe we should have a removed from favorites as well? thoughts @acheng96

Eatery/Eat Now/Detail View Controller/MenuViewController.swift
```diff
@@ -277,6 +287,15 @@ class MenuViewController: UIViewController, MenuButtonsDelegate, TabbedPageViewC
     
     func favoriteButtonPressed() {
         delegate?.favoriteButtonPressed()
+        if eatery.favorite == true{
```
anniexcheng, line 290: Just user eatery.favorite to avoid redundancy

anniexcheng:
I agree with Steven. @myo3, could you add in a 'Removed from Favorites' popup when a eatery is removed from favorites?

say25:
This would probably have the unfilled white star and "Removed from favorites" as text

myo3:
@acheng96 @say25 alright I'll get on it!

lucasderraugh:
Sorry, didn't realize this was still under operation. @myo3 could you start any new work off the new master branch? I made changes that move the logic from the VC into the view and I'd like to avoid display logic being thrown into the VC.

myo3:
Yeah it's not a problem

PR #86, opened 2016-04-24
Kevin/default image

kevingreer:
Uses a default image in the grid vc and the menu vc if the particular eatery doesn't have an image associated with it.

I used the set_image function for the menu header. I'm not sure if that function is what we want, or if we want to just set the image to the raw uiimage.


lucasderraugh:
👍

lucasderraugh:
Merged in c090749

Issue #59, opened 2016-03-20
Drop-down Sorting Menu


anniexcheng:
Clicking on the sort button will animate a drop down menu and clicking on an option in the menu will make the drop down menu disappear:

![image](https://cloud.githubusercontent.com/assets/8130751/14585332/197df0f8-043e-11e6-82a3-1b93ac5fdb5a.png)

say25:
Closed with 23016de

PR #92, opened 2016-05-01
Emily/sort button2

mlelien:
#59 Made drop-down sorting menu


lucasderraugh:
@mlelien There is still a bug in this where upon application launch it will launch the sort options regardless of where I tap. Also, can we try moving some of the setup logic out of the view controller. Let's leave the VC just for presenting the actual view (or converting to presenting a view controller instead of a view).

PR #95, opened 2016-05-07
Add simple brb feature

dennisfedorko:
It's not complete, but here it is...


Eatery/BRBConnectionHandler.swift
```diff
@@ -0,0 +1,169 @@
...
+    
+    /**
+     
+     - Loads login web page
+     
+     */
+    func handleLogin() {
+        loginCount = 0
+        stage = .LoginScreen
+        let loginURL = NSURL(string: "https://get.cbord.com/cornell/full/login.php")!
```
lucasderraugh: Use your above string, also make them let not var

Eatery/BRBConnectionHandler.swift
```diff
@@ -0,0 +1,169 @@
...
+        case LoginFailed
+        case Transition
+        case FundsHome
+        case DiningHistory
+        case Finished
+    }
+    
+    var stage: Stages = .LoginScreen
+    var accountBalance: AccountBalance!
+    var loginURL = "https://get.cbord.com/cornell/full/login.php"
```
lucasderraugh: Basically get rid of this variable if it's only local to that one function

Eatery/BRBConnectionHandler.swift
```diff
@@ -0,0 +1,169 @@
...
+        do {
+            let regex = try NSRegularExpression(pattern: regexString as String, options: .UseUnicodeWordBoundaries)
+            if let match = regex.firstMatchInString(str as String, options: NSMatchingOptions.WithTransparentBounds , range: NSMakeRange(0, str.length)) {
+                return str.substringWithRange(match.rangeAtIndex(0)) as String
+            }
+        } catch {
+            return ""
+        }
+        return ""
+    }
```
lucasderraugh, line 127: Let's convert this to the following since you're not handling any exceptions:

``` swift
let regex = try? NSRegularExpression(pattern: regexString as String, options: .UseUnicodeWordBoundaries)
if let match = regex?.firstMatchInString(str as String, options: NSMatchingOptions.WithTransparentBounds , range: NSMakeRange(0, str.length)) {
    return str.substringWithRange(match.rangeAtIndex(0)) as String
}
return ""
```

Eatery/BRBViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        if connectionHandler.accountBalance != nil && connectionHandler.accountBalance.brbs != "" {
+            timer.invalidate()
+            finishedLogin()
+        }
+    }
+    
+    override func viewDidAppear(animated: Bool) {
+        super.viewDidAppear(animated)
+
+    }
```
lucasderraugh: Redundant

Eatery/BRBViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+            print(self.connectionHandler.stage)
+            if self.connectionHandler.stage == .LoginFailed {
+                self.failedToLogin("incorrect netid and/or password")
+            } else if self.connectionHandler.stage == .LoginScreen {
+                self.login()
+            } else if self.connectionHandler.stage == .FundsHome {
+                self.connectionHandler.getAccountBalance()
+            } else if self.connectionHandler.stage == .DiningHistory {
+                self.connectionHandler.getDiningHistory()
+            }
```
lucasderraugh: If we're testing against enum types, let's use a switch statement.

Eatery/KeychainItemWrapper.swift
```diff
@@ -0,0 +1,175 @@
...
+            
+            if TARGET_IPHONE_SIMULATOR == 1 {
+                tempCheck.removeValueForKey(kSecAttrAccessGroup)
+            }
+            
+            result = SecItemUpdate(updateItem! as CFDictionary, tempCheck as CFDictionary)
+            assert(result == noErr, "Failed to update keychain item")
+        }
+    }
+}
```
lucasderraugh, line 175: Can we Podfile this?

lucasderraugh:
Other than those things, looks pretty good. If you can fixup the random places where there are extra spaces between your last statement and the closing brace and more than 1 space between statements then we're good to go. Great feature!

lucasderraugh:
👍

PR #135, opened 2016-12-01
Daniel/bugfixes

dantheli:
A lot of bug fixes and issue resolutions

### DIFF
--- a/Eatery/Controllers/FilterBar.swift
+++ b/Eatery/Controllers/FilterBar.swift
@@ -101,6 +101,20 @@ class FilterBar: UIView {
         sender.isSelected = !sender.isSelected
         if sender.isSelected {
             selectedFilters.insert(filters[sender.tag])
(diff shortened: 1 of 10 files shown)


Eatery/Controllers/FilterBar.swift
```diff
@@ -27,6 +27,7 @@ fileprivate let filters: [Filter] = [
 ]
 
 protocol FilterBarDelegate: class {
+    var filters: Set<Filter> { get set }
```
dennisfedorko, line 30: not sure why you're using a set here instead of an array, but ok

dennisfedorko:
looks good merge it

Eatery/Controllers/FilterBar.swift
dantheli: It's like marginally faster isn't it
I guess I just wanted to use it :P

PR #147, opened 2017-09-21
Fixed filter view to persist data

kevinchan159:
Fixed the filter view to have the data persist.

The issue before was that didMoveToSuperView was being called before layoutSubviews so no buttons array had length 0. I fixed by creating new function layoutButtons.


Eatery/Controllers/FilterBar.swift
```diff
@@ -63,26 +66,33 @@ class FilterBar: UIView {
...
             button.tag = index
             button.setBackgroundImage(UIImage.image(withColor: UIColor.white), for: .normal)
             button.setBackgroundImage(UIImage.image(withColor: UIColor.eateryBlue), for: .highlighted)
             button.setBackgroundImage(UIImage.image(withColor: UIColor.eateryBlue), for: .selected)
             button.setTitleColor(UIColor.white, for: .selected)
             button.addTarget(self, action: #selector(buttonPressed(sender:)), for: .touchUpInside)
-
+            
             scrollView.addSubview(button)
+            button.snp.makeConstraints({ (make) in
```
dantheli: swiftier to make it a trailing closure
i.e. `button.snp.makeConstraints { make in`

dantheli (approved):
A little petty but next time name your branch something like
`kevin/filter-persist`
So include your name and use dashes instead of underscores :)
Go ahead and merge after fixing the one closure

PR #183, opened 2019-02-01
Ethan/unified eatery status

usubzero:
Made the eatery status code more centralized and updated winter app icon


Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -41,6 +41,13 @@ class EateriesViewController: UIViewController, MenuButtonsDelegate, CLLocationM
     }()
     
     fileprivate var updateTimer: Timer?
+    
+    enum EateryUserStatus: String {
```
whoiswillma: Please push back if you've thought about this, but I think this enum should not have raw values, and instead should have a synthesized field with each of the strings as cases

```
var description: String { 
    switch self {
        // ...
    }
}
```

This doesn't feel like a one-to-one mapping from cases to strings and vice versa, as RawRepresentable would suggest

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -504,24 +511,24 @@ extension EateriesViewController: UICollectionViewDataSource {
         let (section, eateries) = data(for: indexPath.section)
 
         switch section {
-        case "Favorites":
-            header.title = "Favorites"
+        case .favorites:
+            header.title = section.rawValue
```
whoiswillma: Factor `header.title = section.rawValue` out since it's common to all the cases

Eatery/Controllers/Eateries/MenuViewController.swift
```diff
@@ -132,25 +132,17 @@ class MenuViewController: UIViewController, UIScrollViewDelegate, MenuButtonsDel
...
-
-            statusLabel.textColor = .eateryRed
-
-        case let .closed(status, message):
-            statusLabel.text = status
-            hoursLabel.text = message
-
+        case let .closed(message), let .opening(message):
             statusLabel.textColor = .darkGray
+            hoursLabel.text = message
```
whoiswillma: Factor `hoursLabel.text = message` out as well

Eatery/Extensions/Eatery+Extension.swift
```diff
@@ -3,9 +3,41 @@ import SwiftyJSON
...
+        case .opening:
+            return "Opening"
+        case .closing:
+            return "Closing"
+        case .closed:
+            return "Closed"
+        }
+    }
+    
+    var detailString: String {
```
whoiswillma: Nitpick: rename this to `detail` or `message`; Xcode will tell the type if needed

Eatery/Extensions/Eatery+Extension.swift
```diff
@@ -3,9 +3,41 @@ import SwiftyJSON
...
+        case .open:
+            return .eateryGreen
+        case .opening, .closing:
+            return .orange
+        case .closed:
+            return .eateryRed
+        }
+    }
+    
+    var statusString: String {
```
whoiswillma: I'm 50/50 on this one, maybe rename this to `status`? `status`'s are normally strings, but on the other hand `statusColor` is there as well. Do what you feel is best.

Eatery/Extensions/Eatery+Extension.swift
```diff
@@ -86,28 +118,29 @@ extension Eatery {
...
         if isOpenToday() {
             guard let activeEvent = activeEventForDate(Date()) else {
-                return .closed("Closed", "")
+                return .closed("")
             }
-
+            
+            //let eventInfo = (activeEvent.occurringOnDate(Date())) ? (30, "Closing", activeEvent.endDate, ) // TODO-ETHAN
             if activeEvent.occurringOnDate(Date()) {
                 let minutesTillClose = (Int)(activeEvent.endDate.timeIntervalSinceNow/Double(60))
```
whoiswillma, line 125: Change this so it's clear that this is calling the Int initializer. Also, either remove the comment or make it clear why it is there

whoiswillma (approved):
Looks great. Lots of very small things to fix. Let's get that winter icon out

Eatery/Controllers/Eateries/MenuViewController.swift
usubzero: NVM

PR #185, opened 2019-02-08
Removed all nibs from Eatery

whoiswillma:
Nibs are gone from Eatery 🥳


usubzero (commented):
Looks very good overall. Two small nitpicks:


For these protocol declarations;
protocol EateryHeaderCellDelegate: class, protocol EateryMenuCellDelegate: class, protocol FilterDateViewDelegate: class, protocol FilterEateriesViewDelegate: class, protocol MenuButtonsDelegate: class
Why class? Could be better represented by inheriting from AnyObject



In MenuButtonsDelegate, up to you, but I would reconsider removing the optional requirement of favorite button pressed. EateriesViewController doesn't make use of the parameter specified in the requirement, and I just feel it doesn't follow convention to force a function to have unused parameters.

whoiswillma:
screen shot 2019-02-08 at 17 01 47

whoiswillma:
For `MenuButtonsDelegate`, it's standard to pass in the object to a delegate method. Apple uses this throughout their protocols. 

And I changed to use `AnyObject` because `class` protocol inheritance has been deprecated and I was unaware.

usubzero:
> For `MenuButtonsDelegate`, it's standard to pass in the object to a delegate method. Apple uses this throughout their protocols.

That I understand, what I'm wondering is why the the requirement for favoriteButtonPressed was changed from optional to fully required. For example, in EateriesViewController:

`func favoriteButtonPressed(on view: MenuHeaderView) {
	        processEateries()
	        collectionView.reloadData()
	    }`

this function does not make use of the view parameter. Removing the optional requirement makes it such that this parameter is required unnecessarily. In the end it's your call, but just out of curiosity, why was the optional modifier removed?

whoiswillma:
For one method protocols, the function does not need to be optional, since the implementing class can just not inherit from the protocol; it wasn't going to use any other requirements otherwise.

For protocols with more than one method, having the protocol be `@objc` might have some merit, but generally feels unswifty. For Eatery in particular, since all the view controllers that implement the delegate already implement all the methods of that delegate, there's no reason to have them be optional. So, in general, what's the solution here? Closures or some sort of observer framework would be better, although `@objc` optional methods may have some merit for swift code interfacing with Objective-C (which we have none of in Eatery)

Another solution would be to provide an empty default implementation of the protocol methods, but since all these delegates have one or two requirements, I think it just adds clutter (for literally no benefit)

usubzero:
Thanks for explaining your reasoning; makes sense why you chose to remove the optional modified then! Looks good.

PR #186, opened 2019-02-16
Refactor LookAheadViewController

whoiswillma:
- Add activity indicator while loading


usubzero (approved):
Looks good; I appreciate the more appropriate definitions of the utility functions; only doubt I have is in having computed properties for retrieving and setting values of so many view elements in the LookAheadTableViewCell particularly. Feel like it adds a lot of unnecessary code and that it makes referring to these computed properties necessary to see what's actually being changed when reading through the code

PR #188, opened 2019-03-02
William/look ahead design changes

whoiswillma:


chalo2000 (commented):
This looks much more readable! I was wondering how come you started to add .data after some variables since the old code didn't have it. Is it for readability or did you add something that would break the code without the .data?

whoiswillma:
> Is it for readability or did you add something that would break the code without the .data?

The latter; The `Menu` struct is now a container for what used to be menu dictionaries.

PR #190, opened 2019-03-11
Gonzalo/giving day

chalo2000:
I tested the date logic to ensure it only appears once on giving day and that it will not appear before or after.


Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -210,6 +211,25 @@ class EateriesViewController: UIViewController, MenuButtonsDelegate, CLLocationM
         view.addGestureRecognizer(collectionView.panGestureRecognizer)
     }
     
+    func presentGivingDayPopup(){
+        let todayDate = Date()
+        let givingDayStart : Date = {
+            let formatter = DateFormatter()
+            formatter.dateFormat = "MM/dd/yy HH:mm"
+            return formatter.date(from: "03/14/19 00:00")!
```
AAAstorga: this ! hurts me. Is there any way to unwrap this safely. If you can't parse the dates you can just return out of the function?

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -210,6 +211,25 @@ class EateriesViewController: UIViewController, MenuButtonsDelegate, CLLocationM
...
+        }()
+        let givingDayEnd : Date = {
+            let formatter = DateFormatter()
+            formatter.dateFormat = "MM/dd/yy HH:mm"
+            return formatter.date(from: "03/14/19 23:59")!
```
AAAstorga: same

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -210,6 +211,25 @@ class EateriesViewController: UIViewController, MenuButtonsDelegate, CLLocationM
...
+        }()
+        
+        if(UserDefaults.standard.bool(forKey: "didDonatePopup") == false && todayDate > givingDayStart && todayDate < givingDayEnd){
```
AAAstorga: you can do `!UserDefaults.standard.bool....`. No need to do == false

usubzero (approved):

Eatery/Controllers/Eateries/GivingDayViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+
+        //blur effect setup
+        if !UIAccessibility.isReduceTransparencyEnabled {
+            view.backgroundColor = .clear
+            view.isOpaque = false
+            let blurEffect = UIBlurEffect(style: .dark)
+            let blurEffectView = UIVisualEffectView(effect: blurEffect)
+            blurEffectView.frame = self.view.bounds
+            blurEffectView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+
```
AAAstorga: delete space

Eatery/Controllers/Eateries/GivingDayViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+            let blurEffectView = UIVisualEffectView(effect: blurEffect)
+            blurEffectView.frame = self.view.bounds
+            blurEffectView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+            
+            view.addSubview(blurEffectView)
+        } else {
+            view.backgroundColor = .clear
+        }
+        
+        //view setup
```
AAAstorga: space after //

AAAstorga: actually you can just delete the comment, its clear what you are doing

AAAstorga (approved):
Great work! Just small issues. Be sure to rebase!

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: We should add the timepiece pod; It'll be good here, and after backend finishes the refactor we'll need it anyways.

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: Remove parens

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -210,6 +211,25 @@ class EateriesViewController: UIViewController, MenuButtonsDelegate, CLLocationM
...
+            UserDefaults.standard.set(true, forKey: "didDonatePopup")
+                    present(GivingDayViewController(), animated: true, completion: nil)
```
whoiswillma: Unindent

Eatery/Controllers/Eateries/GivingDayViewController.swift
whoiswillma: Constraints for the blur view?

whoiswillma (approved):

Eatery/Controllers/Eateries/EateriesViewController.swift
chalo2000: > We should add the timepiece pod; It'll be good here, and after backend finishes the refactor we'll need it anyways.

I spoke with Young & Daniel and they suggested to always try and minimize the amount of dependencies that our apps have bc if the Timepiece pod goes down, Eatery goes down with it. Since I was really easily able to do this, I don't think it is crucial to use timepiece.

whoiswillma: We can use the Calendar APIs otherwise. 

https://developer.apple.com/documentation/foundation/nscalendar/1407609-date

whoiswillma: Or, pull out the date fomatter from the closures and parse the dates from the strings you have now, which will also let you guard for nil values.

Eatery/Controllers/Eateries/GivingDayViewController.swift
chalo2000: > Constraints for the blur view?

The constraints for the blur view are the bounds of the view (line 24). I was unable to make constraints using Snapkit, so that will have to do.

whoiswillma: Why couldn't you make constraints with Snapkit?

chalo2000: Attempting to use the constraints results in a
Thread 1: Fatal error: Unexpectedly found nil while unwrapping an Optional value

chalo2000: It was stuck inside the scope of the if statement. Setting it as a global variable allowed me to properly make constraints.

PR #199, opened 2019-08-27
William/scroll performance

whoiswillma:
- The search bar and filter bar are in their own section, instead of being added directly to the `collectionView`.
  * This was done to make the search and filter bar stop "jiggling" when scrolling the `collectionView`.
- The AppDev logo reached by pulling down on the `collectionView` is now instantiated in the `EateriesSharedViewController` to stop each `EateryViewController` from needing to add their own copy of it.


Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -209,56 +196,47 @@ class EateriesViewController: UIViewController {
...
-
-        let barsContainerView = UIView()
-        barsContainerView.backgroundColor = nil
-        collectionView.addSubview(barsContainerView)
-        barsContainerView.snp.makeConstraints { make in
-            make.centerX.width.equalToSuperview()
-        }
-
+        // search bar
+
```
young-k, line 208: I don't think we should have `\n` after a comment, I don't think it adds any clarity!

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -372,36 +350,29 @@ class EateriesViewController: UIViewController {
...
+        }
+        
+        var delay: TimeInterval = 0.15
+        for view in views {
+            delay += 0.08
+            UIView.animate(withDuration: 0.55, delay: delay, usingSpringWithDamping: 1.0, initialSpringVelocity: 1.0, options: [.allowUserInteraction], animations: {
+                view.transform = .identity
+                view.alpha = 1.0
+            }, completion: nil)
+        }
```
young-k, line 361: Can't this `for` loop and the above `for view in views` be combined?

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -489,56 +466,80 @@ extension EateriesViewController {
...
+        return 1 + presentedGroups.count
     }
 
     func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        return eateries(in: section).count
+        if section == 0 {
+            return 1
+        } else {
+            return eateries(in: section).count
+        }
```
young-k, line 468: might be cleaner with a ternary, but this is personal preference –– will leave up to you!

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -489,56 +466,80 @@ extension EateriesViewController {
...
-            header.titleLabel.text = "Open"
-            header.titleLabel.textColor = .eateryBlue
-        case .closed:
-            header.titleLabel.text = "Closed"
-            header.titleLabel.textColor = .gray
+        if indexPath.section == 0 {
+            // This header should never be called since
+            // collectionView(layout:referenceSizeForHeaderInSection) should
+            // return CGSize.zero for section 0, but to be safe an empty header
+            // view is provided
```
young-k: Can't this comment fit in two lines // is there a reason for splitting it up?

whoiswillma: I use 80 columns as a soft guide, but I'm up for bumping that to 120

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: Certainly, but I think two for-loops reads better because they distinguish the setup from the animations.

Eatery/Controllers/Eateries/EateriesViewController.swift
young-k: cc @kevinchan159

whoiswillma:
I noticed an issue where entering search text would cause the keyboard to dismiss. Because `reloadData` was being called in the `reloadEateries` method, the search bar would be removed from the window and then reattached, causing it to lose its first responder status. 

Instead of trying to track which sections to insert and remove, I made the number of sections constant and just hide and show them as needed, which is simpler and also stops the search and filter bar cell from being reloaded

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -372,36 +336,29 @@ class EateriesViewController: UIViewController {
...
+        collectionView.layoutIfNeeded()
+        collectionView.isHidden = false
+        
+        let cells = collectionView.visibleCells as [UIView]
+        let headers = collectionView.visibleSupplementaryViews(ofKind: UICollectionElementKindSectionHeader) as [UIView]
+        let views = (cells + headers).sorted {
+            $0.convert($0.bounds, to: nil).minY < $1.convert($1.bounds, to: nil).minY
+        }
+        
+        for view in views {
```
kevinchan159, line 349: nit: pref to use `forEach`

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -372,36 +336,29 @@ class EateriesViewController: UIViewController {
...
+            view.transform = CGAffineTransform(translationX: 0.0, y: 32.0)
+            view.alpha = 0.0
+        }
+        
+        var delay: TimeInterval = 0.15
+        for view in views {
```
kevinchan159, line 355: same here

Eatery/Controllers/Eateries/EateriesViewController.swift
kevinchan159: i agree w/ young here. you could just add comments inside the for loop but if you feel strongly about this you can keep it

Eatery/Controllers/Eateries/EateriesViewController.swift
kevinchan159: pref ternary here too

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -586,16 +578,41 @@ extension EateriesViewController: UICollectionViewDelegate {
...
 extension EateriesViewController: UICollectionViewDelegateFlowLayout {
-
+ 
+    func collectionView(_ collectionView: UICollectionView,
+                        layout collectionViewLayout: UICollectionViewLayout,
+                        sizeForItemAt indexPath: IndexPath) -> CGSize {
+        if indexPath.section == 0 {
+            let height = searchFilterContainer.systemLayoutSizeFitting(UILayoutFittingCompressedSize).height
+            return CGSize(width: collectionView.bounds.width, height: height)
+        } else {
```
kevinchan159: nit: could remove the `else` since you're returning on line 590

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -586,16 +578,41 @@ extension EateriesViewController: UICollectionViewDelegate {
...
+            return gridLayout.itemSize
+        }
+    }
+    
+    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, insetForSectionAt section: Int) -> UIEdgeInsets {
+        if section == 0 || eateries(in: section).isEmpty {
+            return .zero
+        } else {
+            return gridLayout.sectionInset
```
kevinchan159: same here

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -586,16 +578,41 @@ extension EateriesViewController: UICollectionViewDelegate {
...
+    
     func collectionView(_ collectionView: UICollectionView,
                         layout collectionViewLayout: UICollectionViewLayout,
                         referenceSizeForHeaderInSection section: Int) -> CGSize {
-        return CGSize(width: collectionView.frame.width, height: 56.0)
+        if section == 0 || eateries(in: section).isEmpty {
+            // Returning CGSize.zero here causes the collectionView to not query for a header at all, but
+            // collectionView(_:viewForSupplementaryElementOfKind:at:) returns a blank header to be safe
+            return .zero
+        } else {
```
kevinchan159: same here

kevinchan159 (changes requested):
looks great! just left some small nits

Eatery/Controllers/Eateries/EateriesViewController.swift
kevinchan159: Lets go w/ 100 characters as the char limit per line. Thats the recommended limit in the Google/Airbnb style guides

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: Why in general is it preferable to use `forEach` over the regular `for` statements?

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: I don't feel particularly strong about this; It's more a case of it was that way before (which is not a good reason in-and-of-itself, but which I think is reasonable if I came to the code like it is) in combination with (what I perceive as) the additional readability.

It's more so that the `for`-loops behave as separate paragraphs in an essay: they serve different purposes, therefore they should be split, even if semantically having one and having two are the same.

Eatery/Controllers/Eateries/EateriesViewController.swift
kevinchan159: There isnt really a benefit of using one over the other besides consistency. Since Swift is functional and we try to use functional functions like `map/filter/reduce` where possible, I think using `forEach` where possible makes sense

kevinchan159 (approved):
approving to unblock but address comments before merging pls!

kevinchan159:
small note: Could you click on `Resolve conversation` after you finish addressing a comment? Just makes it a bit easier for reviewers to know what you've addressed

Eatery/Controllers/Eateries/EateriesCollectionViewGridLayout.swift
```diff
@@ -0,0 +1,40 @@
...
+    
+    override func prepare() {
+        guard let collectionView = collectionView else {
+            return
+        }
+        
+        let margin = EateriesViewController.collectionViewMargin
+        
+        let cellWidth: CGFloat
+        if collectionView.traitCollection.horizontalSizeClass == .regular
```
chalo2000, line 25: Perhaps consider replacing this with a ternary

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -180,96 +159,98 @@ class EateriesViewController: UIViewController {
...
 
-        refreshControl.tintColor = .white
-        refreshControl.addTarget(self, action: #selector(refreshEateries), for: .valueChanged)
+        activityIndicator.startAnimating()
+        failedToLoadView.alpha = 0
 
-        collectionView.refreshControl = refreshControl
+        updateTimer = Timer.scheduledTimer(withTimeInterval: 60, repeats: true) { [weak self] _ in
+            guard let `self` = self else { return }
+            print("Updating \(type(of: self))", Date())
```
chalo2000, line 179: Remove print statement

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -489,56 +459,76 @@ extension EateriesViewController {
 extension EateriesViewController: UICollectionViewDataSource {
 
     func numberOfSections(in collectionView: UICollectionView) -> Int {
-        return presentedGroups.count
+        return 4
```
chalo2000, line 462: Why are there 4 sections if our switch above only has cases for 3 sections and we return a 1 whenever section == 0  below? Something seems off to me with this.

chalo2000 (commented):
Just tiny fixes and a question about the sections in the collectionView

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: I see three main reasons for sticking with "regular" `for` loops with Eatery:
1. Eatery already uses `for` almost everywhere, so being consistent is good (when it makes sense)
2. `forEach` produces `Void` as its value, whereas `map` (and the other functional methods in Swift) produce meaningful values. This matters because `for` loops are control flow statements, and should therefore be distinguished from expressions that produce values. Instead of having to read the words `forEach`, `for` loops are separated visually from the rest of the code, by the syntax itself as well as by syntax highlighting. 
3. `for` loops are generally used to manipulate state, whereas `map` is expected to _not_  manipulate state. Similarly to 2., as human beings who have to look at code with our eye-balls, having this separation is useful for distinguishing the purpose of code.

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: Hmmn, I feel keeping this one is worthwhile. It doesn't get printed that often and it's also a pretty significant event.

Eatery/Controllers/Eateries/EateriesViewController.swift
whoiswillma: The search and filter bars are now a cell in the collection view. In order to make Some Things™ easier, I put them in their own section. Now there are four sections: the "bars" section which only ever contains one cell, and then the remaining three sections ("favorites", "open", and "closed") as usual.

Eatery/Controllers/Eateries/EateriesCollectionViewGridLayout.swift
whoiswillma: I think a ternary would be a bit too unwieldy: the conditional and both expressions are pretty long.

chalo2000 (approved):

Issue #202, opened 2019-09-02
Refactor Meal Plan Login and Account into separate view controllers


whoiswillma:
The meal plan view controller currently switches between the login view and account info view by changing the opacity of views. 

We should instead use more view controllers to effectively manage the presentation and loading states of the login and accounts views.

PR #206, opened 2019-09-02
Refactor BRB Login by separating the login view from the accounts view

whoiswillma:
Issue: #202 

- Separate the login view from the accounts view by making the login view into a view controller, and adding an additional view controller `BRBViewController` to manage the `BRBLoginViewController` and `BRBAccountViewController`. 
  * This was previously accomplished by adjusting the view alpha on each view. 
- Login View Changes:
   * Add new error view
   * Use a `UISwitch` to save login info

Simulator Screen Shot - iPhone 8 - 2019-09-02 at 14 06 43


Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+//  BRBAccountViewController.swift
+//  Eatery
+//
+//  Created by William Ma on 9/1/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import UIKit
+
+// MARK: -
```
kevinchan159: delete if unused

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        tableView = UITableView(frame: .zero, style: .grouped)
+        tableView.register(BRBBalanceTableViewCell.self, forCellReuseIdentifier: "balance")
+        tableView.register(BRBHistoryTableViewCell.self, forCellReuseIdentifier: "history")
+        tableView.allowsSelection = false
+        tableView.separatorColor = .wash
+        tableView.dataSource = self
+        tableView.delegate = self
+        view.addSubview(tableView)
+        tableView.snp.makeConstraints { make in
+            make.top.edges.equalToSuperview()
```
kevinchan159: I believe doing `edges` already includes `top` so you prob just need `make.edges.equalToSuperview()`

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+}
+
+// MARK: -
```
kevinchan159: Delete if unused or add some sort of description

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+
+extension BRBAccountViewController: UITableViewDataSource {
+    
+    func numberOfSections(in tableView: UITableView) -> Int {
+        return 2
+    }
+    
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        if section == 0 {
+            return 4
```
kevinchan159, line 58: Maybe add a comment for what the 4 sections are.

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return account.history.count
+    }
+    
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        if indexPath.section == 0 {
+            let cell = tableView.dequeueReusableCell(withIdentifier: "balance") as! BRBBalanceTableViewCell
```
kevinchan159: Can we make the cell identifiers a constant since we use it in more than one place?

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+
+            switch indexPath.row {
+            case 0:
+                cell.titleLabel.text = " Meal Swipes"
+                let swipesLeft = account.swipes == "" ? "Unlimited" : account.swipes
+                cell.subtitleLabel.text = swipesLeft
+            case 1:
+                cell.titleLabel.text = " BRBs"
+                cell.subtitleLabel.text = "$" + account.brbs
```
kevinchan159: nit: `"$(account.brbs)"`

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            case 2:
+                cell.titleLabel.text = " Laundry"
+                cell.subtitleLabel.text = "$" + account.laundry
+            case 3:
+                cell.titleLabel.text = " City Bucks"
+                cell.subtitleLabel.text = "$" + account.cityBucks
```
kevinchan159: same here and for `case 2:` as well

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            default:
+                break
+            }
+            
+            return cell
+        } else {
```
kevinchan159: you can get rid of `else` since you are returning on line 84

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            let cell = tableView.dequeueReusableCell(withIdentifier: "history") as! BRBHistoryTableViewCell
+
+            let history = account.history[indexPath.row ]
+            cell.titleLabel.text = history.name
+            cell.subtitleLabel.text = history.timestamp
+
+            cell.priceLabel.text = "-$" + history.amount
```
kevinchan159: same here

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -0,0 +1,338 @@
...
+//  Eatery
+//
+//  Created by William Ma on 9/1/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import NVActivityIndicatorView
+import UIKit
+
+// MARK: -
```
kevinchan159: Add description or delete

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -0,0 +1,338 @@
...
+        keychain["password"] = nil
+        
+        BRBAccountSettings.saveLoginInfo = false
+    }
+    
+    static func loadFromKeychain() -> LoginInfo? {
+        let keychain = KeychainItemWrapper(identifier: "netid", accessGroup: nil)
+        if let netid = keychain["netid"] as? String, let password = keychain["password"] as? String {
+            return (netid: netid, password: password)
+        } else{
```
kevinchan159: remove the `else` since you have `return` on line 46

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -0,0 +1,338 @@
...
+                             didRequestLoginWithNetid netid: String,
+                             password: String)
+    
+}
+
+// MARK: - BRB Login View Controller
+
+class BRBLoginViewController: UIViewController {
+    
+    weak var delegate: BRBLoginViewControllerDelegate?
```
kevinchan159, line 61: can we make this private?

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -0,0 +1,338 @@
...
+        if animated {
+            UIViewPropertyAnimator(duration: 0.35, dampingRatio: 1, animations: actions).startAnimation()
+        } else {
+            actions()
+        }
+    }
+    
+}
+
+// MARK: -
```
kevinchan159: delete or add description

Eatery/Controllers/BRB/BRBPrivacyStatementViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+//  BRBPrivacyStatementViewController.swift
+//  Eatery
+//
+//  Created by William Ma on 9/1/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import UIKit
+
+// MARK: -
```
kevinchan159: delete or add description

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -1,285 +1,167 @@
...
+        state = newState
     }
 
 }
 
-extension BRBViewController: BRBLoginViewDelegate {
-
-    func brbLoginViewClickedLogin(brbLoginView: BRBLoginView, netid: String, password: String) {
-        requestStart = Date()
+// MARK: -
```
kevinchan159: delete or add description

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -1,285 +1,167 @@
...
         connectionHandler.netid = netid
         connectionHandler.password = password
         connectionHandler.handleLogin()
+        
+        loginViewController.isLoading = true
+        
+        requestStart = Date()
     }
-
+
```
kevinchan159, line 113: delete whitespace

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -1,285 +1,167 @@
...
+        loginViewController.isLoading = true
+        
+        requestStart = Date()
     }
-
+    
 }
 
-extension BRBViewController: BRBConnectionDelegate {
+// MARK: -
```
kevinchan159: delete or add description

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -1,285 +1,167 @@
...
-
-        addLoginView()
-
-        loginView?.loginFailedWithError(error: error)
+        loginViewController.errorDescription = error
+        loginViewController.setShowErrorMessage(true, animated: true)
+        loginViewController.isLoading = false
     }
-
+
```
kevinchan159, line 145: delete whitespace

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -1,285 +1,167 @@
...
+        loginViewController.errorDescription = error
+        loginViewController.setShowErrorMessage(true, animated: true)
+        loginViewController.isLoading = false
     }
-
+    
 }
 
-extension BRBViewController: SettingsTableViewControllerDelegate {
+// MARK: -
```
kevinchan159: delete or add description

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -1,285 +1,167 @@
...
-        addLoginView()
-
-        //Need to add to subview so it works on an actual device
-        view.addSubview(connectionHandler)
-
+        
         navigationController?.popToViewController(self, animated: true)
     }
-
+
```
kevinchan159, line 158: delete whitespace

Eatery/Views/BRB/BRBLoginErrorView.swift
```diff
@@ -0,0 +1,58 @@
...
+//  Eatery
+//
+//  Created by William Ma on 9/2/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import SnapKit
+import UIKit
+
+// MARK: -
```
kevinchan159: delete or add description

Eatery/Controllers/BRB/BRBAccountViewController.swift
chalo2000: Is this intentional or did you forget to put something here?

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+                cell.subtitleLabel.text = "$" + account.cityBucks
+            default:
+                break
+            }
+            
+            return cell
+        } else {
+            let cell = tableView.dequeueReusableCell(withIdentifier: "history") as! BRBHistoryTableViewCell
+
+            let history = account.history[indexPath.row ]
```
chalo2000:
```suggestion
            let history = account.history[indexPath.row]
```

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            cell.priceLabel.text = "-$" + history.amount
+            cell.priceLabel.textColor = .eateryRed
+            
+            return cell
+        }
+    }
+    
+}
+
+// MARK: -
```
chalo2000: ditto

Eatery/Controllers/BRB/BRBLoginViewController.swift
chalo2000: ditto

kevinchan159 (changes requested):
awesome work!!! 😍 great job following the PR guidelines and adding a screenshot! 
just a small note, it might be better to make the PR's a bit smaller in the future just so that its easier for people to look at past commits to help debug any issues that may arise!

Eatery/Controllers/BRB/BRBLoginViewController.swift
chalo2000: ditto

Eatery/Controllers/BRB/BRBViewController.swift
chalo2000: ditto

Eatery/Controllers/BRB/BRBViewController.swift
chalo2000: ditto

Eatery/Controllers/BRB/BRBViewController.swift
chalo2000: ditto

chalo2000 (changes requested):
Other than minuscule issues, LGTM

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        return account.history.count
+    }
+    
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        if indexPath.section == 0 {
+            let cell = tableView.dequeueReusableCell(withIdentifier: "balance") as! BRBBalanceTableViewCell
+
+            switch indexPath.row {
+            case 0:
+                cell.titleLabel.text = " Meal Swipes"
```
Omarrasheed: Can we move this setup stuff into a cell.configure().

Eatery/Controllers/BRB/BRBAccountViewController.swift
whoiswillma: This was just a habit I got into over the summer. I'll get rid of them to follow the AppDev style guide

Eatery/Controllers/BRB/BRBLoginViewController.swift
whoiswillma: Do you mean make the delegate private? We need write access to it from the `BRBViewController`

Eatery/Controllers/BRB/BRBViewController.swift
whoiswillma: In general should we keep whitespace only lines?

Eatery/Controllers/BRB/BRBAccountViewController.swift
Omarrasheed: You can just store an array of those `titleLabel.text` and `subtitleLabel.text` values at the top and instead of having this switch statement, just directly index that array. That'll make it 10x easier to change the order or make any big structural changes to this stuff down the line.

Could also move the two text params into some struct to ensure they're being paired correctly. This may be overkill tho cc @kevinchan159

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+            }
+            
+            return cell
+        } else {
+            let cell = tableView.dequeueReusableCell(withIdentifier: "history") as! BRBHistoryTableViewCell
+
+            let history = account.history[indexPath.row ]
+            cell.titleLabel.text = history.name
+            cell.subtitleLabel.text = history.timestamp
+
```
Omarrasheed: nit: delete newline

Eatery/Controllers/BRB/BRBLoginViewController.swift
kevinchan159: oh i see. could we pass pass in the `delegate` by creating a custom init for `BRBLoginViewController` and then change the `delegate` to be private? This way, the delegate will be set only once and we prevent the possibility of someone changing it later. We should make all delegates private because other classes shouldn't be accessing them.

Eatery/Controllers/BRB/BRBViewController.swift
kevinchan159: its fine to have newlines if it helps with reading code but i like to delete unnecessary whitespace like in this case

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -0,0 +1,338 @@
...
+            let loginInfo = (netid: netid, password: password)
+            BRBAccountSettings.saveToKeychain(loginInfo: loginInfo)
+        }
+        
+        delegate?.loginViewController(self, didRequestLoginWithNetid: netid, password: password)
+    }
+    
+    func setShowErrorMessage(_ newValue: Bool, animated: Bool) {
+        let actions: () -> Void = { [weak self] in
+            self?.errorView.isCollapsed = !newValue
```
Omarrasheed: can we unwrap self with a guard let?

Eatery/Views/BRB/BRBBalanceTableViewCell.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Created by William Ma on 9/2/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import UIKit
+
+class BRBBalanceTableViewCell: UITableViewCell {
+    
+    private(set) var titleLabel: UILabel!
+    private(set) var subtitleLabel: UILabel!
```
Omarrasheed: This is cool i didn't know this was a thing. Def gonna use this

Eatery/Controllers/BRB/BRBViewController.swift
whoiswillma: What about whitespace between lines of code that are not top-level?

```
class A {
    func x() {
        ...
    }
____< whitespace here?
    func y() {
        ...
    }
}
```

Eatery/Controllers/BRB/BRBAccountViewController.swift
kevinchan159: i think the struct is a bit of an overkill

Eatery/Controllers/BRB/BRBViewController.swift
kevinchan159: i would also prefer to not have whitespace there

whoiswillma: Ok, I've enabled the "Automatically Trim Whitespace only lines" feature in Xcode. There are tons of lines that are whitespace only that I've committed (many many more than are mentioned in PR comments). Instead of going back through and deleting them all individually I'd rather just let them dissipate over time as the files get edited.

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,115 @@
+//
+//  BRBAccountViewController.swift
+//  Eatery
+//
+//  Created by William Ma on 9/1/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
```
young-k, line 6: Could we update this to be `Cornell AppDev`, cc @kevinchan159

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -0,0 +1,329 @@
...
+    private var netidTextField: UITextField!
+    
+    private var passwordPrompt: UILabel!
+    private var passwordTextField: UITextField!
+    
+    private var saveLoginInfoSwitch: UISwitch!
+    
+    private var loginButton: UIButton!
+    private var activityIndicator: NVActivityIndicatorView!
+    var isLoading: Bool = false {
```
young-k, line 85: probably cleaner to have a newline before this one

Eatery/Controllers/BRB/BRBLoginViewController.swift
whoiswillma: Seems like quite a bit of overhead for a pattern that's well established: The only owner of the `BRBLoginViewController`(the `BRBViewController`), has its reference to it private, so there's no chance of it being changed unintentionally outside of the owners file. 

Delegation is a well established pattern within Apple's frameworks, and it (almost) always works by assigning to a weak variable. Having that consistency with Apple's frameworks is also important. 

It's true that other classes shouldn't be accessing delegates, but I think that it's generally understood that delegates should only be assigned to. It seems that someone would have to be intentionally writing malicious code or just be uninformed to access the delegate in that way.

Eatery/Controllers/BRB/BRBViewController.swift
Omarrasheed: swiftlint autocorrect should get all of em for u

Eatery/Controllers/BRB/BRBLoginViewController.swift
whoiswillma: I think the better action here would just be to retain a strong reference to `self`. In the first case, the `UIViewPropertyAnimator` immediately falls out of scope, and in the second case the closure never even leaves the function scope

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,118 @@
...
+            cell.subtitleLabel.text = swipesLeft
+        case 1:
+            cell.titleLabel.text = "BRBs"
+            cell.subtitleLabel.text = "$\(account.brbs)"
+        case 2:
+            cell.titleLabel.text = "Laundry"
+            cell.subtitleLabel.text = "$\(account.laundry)"
+        case 3:
+            cell.titleLabel.text = "City Bucks"
+            cell.subtitleLabel.text = "$\(account.cityBucks)"
```
kevinchan159: can we add a `func configure(with title: string, subtitle:string) {}` function to the cell and just pass these values in there instead?

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -0,0 +1,118 @@
...
+        default:
+            break
+        }
+    }
+
+    private func configure(historyTableViewCell cell: BRBHistoryTableViewCell, for indexPath: IndexPath) {
+        let history = account.history[indexPath.row]
+        cell.titleLabel.text = history.name
+        cell.subtitleLabel.text = history.timestamp
+        cell.priceLabel.text = "-$" + history.amount
```
kevinchan159: `"-$\(history.amount)"`

Eatery/Controllers/BRB/BRBLoginViewController.swift
kevinchan159: Oh I think there was a bit of confusion. I totally agree that the delegate should be `weak`. I was just saying I think setting it to be `private` in addition would help. I also agree that in 99% of the cases, people probably won't write malicious code but I think additional layers of security are always good. The argument is similar for why we set view variables as private. It is understood that we shouldn't be setting these view variables outside of the class but making them `private` just helps to guarantee it. Passing the delegate through the init also guarantees that we only set the delegate once.

kevinchan159 (approved):
approving to unblock but just left 3 more comments!

chalo2000 (approved):

PR #209, opened 2019-09-03
Fix filter view position on notched devices

whoiswillma:
Issue #205 

out


Eatery/Views/Lookahead/FilterEateriesView.swift
```diff
@@ -20,6 +20,8 @@ class FilterEateriesView: UIView, UIGestureRecognizerDelegate {
 
     private let dateAndMealSeparator = UIView()
 
+    private(set) var separatorY: CGFloat = 0
```
kevinchan159, line 23: does this need to be `private(set)`? can we just make it private?

kevinchan159 (approved):
nice work! just 1 question

Eatery/Views/Lookahead/FilterEateriesView.swift
whoiswillma: We need to access it from `LookaheadTableViewController`

Issue #204, opened 2019-09-02
Fix eateries cell awkward pop-up


chalo2000:
When scrolling down, some eateries will play the pop-up animation as if they had just loaded.

PR #210, opened 2019-09-04
Gonzalo/awkward cell popup

chalo2000:
Issue #204 

I went back to William's older commits to find where the bug began and eventually narrowed down the issue to starting within the scrollDidStart function and then to the setUpPill function's use of UIViewPropertyAnimator. Swiping too fast, upon opening the app, would cause cells to instantiate when the pill was playing its animation, causing the cells to share the animation. Using layoutIfNeeded() before the animation begins fixes this issue.

Before:
ezgif com-resize (1)

After:
ezgif com-resize

### DIFF
--- a/Eatery/Controllers/Eateries/PillViewController.swift
+++ b/Eatery/Controllers/Eateries/PillViewController.swift
@@ -101,6 +101,8 @@ class PillViewController: UIViewController {
     }
 
     func setShowPill(_ showPill: Bool, animated: Bool) {
+        // Prevents newly created cells from sharing the pill animation
+        view.layoutIfNeeded()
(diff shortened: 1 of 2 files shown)


lucyxubroad (approved):
Wow, good find! Looks good to me! :)

Eatery/Controllers/Eateries/PillViewController.swift
```diff
@@ -101,6 +101,7 @@ class PillViewController: UIViewController {
     }
 
     func setShowPill(_ showPill: Bool, animated: Bool) {
+        view.layoutIfNeeded() // Prevents newly created cells from sharing the pill animation
```
kevinchan159: can we move the comment to the previous line

kevinchan159 (approved):
nice!

Issue #207, opened 2019-09-03
Eateries appear in the wrong section


whoiswillma:
After leaving the app open for long enough, especially during times when Eateries are opening and closing, eateries will not move from one section to another. 

i.e. If Becker goes from Closed -> Open at 6PM, and the Eatery app is left open from 5:57PM to 6:03PM, Becker will remain in the Closed section despite being Open (and the cell displaying that it is open).

This is caused because `EateriesViewController` does not refresh its sections every minute; only its cells.

PR #213, opened 2019-09-05
Reload eateries every minute

whoiswillma:
Fixes issue: #207

### DIFF
--- a/Eatery/Controllers/Eateries/EateriesViewController.swift
+++ b/Eatery/Controllers/Eateries/EateriesViewController.swift
@@ -60,6 +60,24 @@ protocol EateriesViewControllerScrollDelegate: AnyObject {
 
 // MARK: - Eateries View Controller
 
+/**
+ The `EateriesViewController` manages the presentation of Eateries. It works closely with its data
+ source to control the content it displays.
+
+ This view controller was intended to abstract shared functionality from the
+ `CampusEateriesViewController` and `CollegetownEateriesViewController`. As such it is (somewhat)
+
+ The `EateriesViewController` has three states:
+
+ 1. presenting - eateries are actively presented and the Eateries View Controller can query its data
+ source for updated Eateries whenever
+ 2. loading - a loading indicator is displayed and no eateries are presented. The Eateries View
+ Controller will not query its data source for Eateries in this state.
+ 3. failedToLoad - an error view is displayed and no eateries are presented. The Eateries View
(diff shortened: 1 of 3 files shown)


usubzero (commented):
Looks good; may be a bit of a pick but why not have it reload eateries on 30 minute time intervals of the hour rather than every minute? Eateries don't tend to close at odd times so there's way more reloading than needed

whoiswillma:
> Looks good; may be a bit of a pick but why not have it reload eateries on 30 minute time intervals of the hour rather than every minute? Eateries don't tend to close at odd times so there's way more reloading than needed

Eateries don't normally close at weird times but I think having it work all the time is better than only most of the time. 

Since we have no control over the exact data the backend provides, it's better to be more accommodating than to be more strict. Plus refreshing (only the eateries) is pretty lightweight in the grand scheme of things.

usubzero:
I'm just saying, one minute feels a bit excessive. Eateries status' won't change at xx:26, but it's your call.

usubzero (approved):

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -398,13 +410,32 @@ class EateriesViewController: UIViewController {
...
+            }
+        }
+
+        self.eateriesByGroup = newEateriesByGroup
+
+        if animated {
             collectionView.reloadSections(IndexSet(1...3))
+        } else {
+            UIView.performWithoutAnimation {
+                collectionView.reloadSections(IndexSet(1...3))
```
kevinchan159: maybe extract `IndexSet(1...3)` into a variable so that its clear what it represents and also because we use it twice

kevinchan159 (approved):
😍

chalo2000 (commented):
I was wondering why you included the os_log calls and kept the print statement inside scheduleUpdateTimer. After testing to see that the expected values are printed, wouldn't these just serve as clutter to the console?

whoiswillma:
> I was wondering why you included the os_log calls and kept the print statement inside scheduleUpdateTimer. After testing to see that the expected values are printed, wouldn't these just serve as clutter to the console?

If we never print anything there are no messages to _be_ cluttered up. 

Reloading is a significant event: it could possibly change the contents of every visible cell. It can be helpful to see when such events happen in order to keep track of state beyond what's visible on the presentation layer

Issue #214, opened 2019-09-09
Firebase Analytics Events


chalo2000:
We want to get the number of presses done on
1) Collegetown button on pill
2) Campus button on pill
3) Lookahead tab button
4) BRB Account tab button
5) BRB Login Button
6) Any Campus filter button
7) Any Collegetown filter button

kevinchan159:
should try to use `AppDevAnalytics` pod for this!

chalo2000:
8) Eatery tab button

PR #215, opened 2019-09-15
Add analytics events

chalo2000:
Issue: #214 

Implements the events noted in #214. While events are implemented along with button presses activating them, they are still not showing on Firebase Analytics (TODO).

### DIFF
--- a/Eatery/Analytics.swift
+++ b/Eatery/Analytics.swift
@@ -1,0 +1,73 @@
+//
+//  Analytics.swift
+//  AppDevAnalytics
+//
+//  Created by Kevin Chan on 9/4/19.
+//  Copyright © 2019 Kevin Chan. All rights reserved.
+//
+
+import Crashlytics
+
+class AppDevAnalytics {
+
+    static let shared = AppDevAnalytics()
+
+    private init() {}
+
+    func log(_ payload: Payload) {
+        #if !DEBUG
+        let fabricEvent = payload.convertToFabric()
+        Answers.logCustomEvent(withName: fabricEvent.name, customAttributes: fabricEvent.attributes)
+        #endif
+    }
+}
+
+    // MARK: Event Payloads
+
+    /// Log whenever Collegetown pill button is pressed
+    struct CollegetownPressPayload: Payload {
+        static let eventName: String = "Collegetown Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever Campus pill button is pressed
+    struct CampusPressPayload: Payload {
+        static let eventName: String = "Campus Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever Eatery tab bar button is pressed
+    struct EateryPressPayload: Payload {
+        static let eventName: String = "Eatery Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever LookAhead tab bar button is pressed
+    struct LookAheadPressPayload: Payload {
+        static let eventName: String = "LookAhead Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever BRB Account tab bar button is pressed
+    struct BRBPressPayload: Payload {
+        static let eventName: String = "BRB Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever BRB Account login button is pressed
+    struct BRBLoginPressPayload: Payload {
+        static let eventName: String = "BRB Login Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever Campus filter button is pressed
+    struct CampusFilterPressPayload: Payload {
+        static let eventName: String = "Campus Filter Press"
+        let deviceInfo = DeviceInfo()
+    }
+
+    /// Log whenever Collegetown filter button is pressed
+    struct CollegetownFilterPressPayload: Payload {
+        static let eventName: String = "Collegetown Filter Press"
+        let deviceInfo = DeviceInfo()
+    }
--- a/Eatery/Analytics.swift
+++ b/Eatery/Analytics.swift
@@ -7,19 +7,27 @@
 //
 
 import Crashlytics
-
+import FirebaseAnalytics
 class AppDevAnalytics {
 
     static let shared = AppDevAnalytics()
 
     private init() {}
 
-    func log(_ payload: Payload) {
+    func logFabric(_ payload: Payload) {
         #if !DEBUG
         let fabricEvent = payload.convertToFabric()
         Answers.logCustomEvent(withName: fabricEvent.name, customAttributes: fabricEvent.attributes)
         #endif
     }
+
+    func logFirebase(_ payload: Payload) {
+        #if !DEBUG
+        let firebaseEvent = payload.convertToFirebase()
+        Analytics.logEvent(firebaseEvent, parameters: nil)
+        #endif
+    }
+
 }
 
 // MARK: Event Payloads
--- a/Eatery/Analytics.swift
+++ b/Eatery/Analytics.swift
@@ -8,18 +8,12 @@
 
 import Crashlytics
 import FirebaseAnalytics
+
 class AppDevAnalytics {
 
     static let shared = AppDevAnalytics()
 
     private init() {}
-
-    func logFabric(_ payload: Payload) {
-        #if !DEBUG
-        let fabricEvent = payload.convertToFabric()
-        Answers.logCustomEvent(withName: fabricEvent.name, customAttributes: fabricEvent.attributes)
-        #endif
-    }
 
     func logFirebase(_ payload: Payload) {
         #if !DEBUG
--- a/Eatery/Analytics.swift
+++ b/Eatery/Analytics.swift
@@ -22,52 +22,107 @@ class AppDevAnalytics {
     }
 }
 
-    // MARK: Event Payloads
+// MARK: Event Payloads
 
-    /// Log whenever Collegetown pill button is pressed
-    struct CollegetownPressPayload: Payload {
-        static let eventName: String = "Collegetown Press"
-        let deviceInfo = DeviceInfo()
-    }
+/// Log whenever Collegetown pill button is pressed
+struct CollegetownPressPayload: Payload {
+    static let eventName: String = "collegetown_pill_press"
+    let deviceInfo = DeviceInfo()
+}
 
-    /// Log whenever Campus pill button is pressed
-    struct CampusPressPayload: Payload {
-        static let eventName: String = "Campus Press"
-        let deviceInfo = DeviceInfo()
-    }
+/// Log whenever Campus pill button is pressed
+struct CampusPressPayload: Payload {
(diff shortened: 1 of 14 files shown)


whoiswillma (changes requested):
All of the code looks great! I think I have a few high level (i.e. non-code) suggestions and questions:

1. I think we should move these files into a separate logging source code folder

2. You'll have to forgive me for not having any experience with the AppDevAnalytics framework, but I have two questions: 

    a. Can we move this into a Cocoapod? It's a bit strange to have this seemingly project agnostic code in the Eatery codebase. @kevinchan159 might be better suited to answer this question

    b. Is this PR supposed to be logging events that are actually happening within the app? Or is this just setting up the framework to later do so.

chalo2000:
So the Analytics and EventPayload files are straight from the AppDevAnalytics cocoa pod. I spoke with Kevin and since there is currently an issue with SwiftyJSON, he recommended I just copy the files and later he will fix the issue with the cocoapod and edit it. Thus, when we have the cocoa pod working, the “project agnostic” code will be in its own folder. As for the purpose of this PR, I was just committing what I had because the next step would be getting the events to work with Firebase (so this is a framework).

whoiswillma:
Alrighty! LGTM then.

whoiswillma (approved):

Eatery/Analytics.swift
```diff
@@ -0,0 +1,73 @@
...
+        #if !DEBUG
+        let fabricEvent = payload.convertToFabric()
+        Answers.logCustomEvent(withName: fabricEvent.name, customAttributes: fabricEvent.attributes)
+        #endif
+    }
+}
+
+    // MARK: Event Payloads
+
+    /// Log whenever Collegetown pill button is pressed
```
kevinchan159: why are these indented?

kevinchan159 (changes requested):
nice, one question!

Eatery/Analytics.swift
chalo2000: From transit:

> // MARK: Event Payloads
>
>// MARK: Important
>/// Log app launch with device info
>struct AppLaunchedPayload: Payload {
>    static let eventName: String = "App Launched"
>    let deviceInfo = DeviceInfo()
>}

I thought a MARK: Important was unnecessary since MARK: Event Payloads is more descriptive. Also, I feel like MARKs are “titles” and so I like to give them some indentation instead of tacking on the events right after because it is significant and should be visually separated (so it doesn’t blend in if I’m scrolling through a file looking for something). The same reason applies for why each event has an indentation between them (they’re each significant). I’m assuming you’re asking because you feel the opposite way; how come?

kevinchan159: This reasoning is similar to if you have two classes in the same file. It would look like:
```
class A {

}

class B {

}
```

chalo2000: Right; so what point are you trying to make?

kevinchan159: Wait I might be tripping but right now I see your changes looks like:
```
    struct A {
    }
    
    struct B {
    }
```
when it should look like
```
struct A {
}
    
struct B {
}
```

chalo2000: Nope I’m tripping. That probably happened bc I copy pasted it in from a previous branch. Don’t understand why it always tends to indent every time you paste. I’ll fix it ASAP

chalo2000:
I fixed the indents and I also added some more events and renamed older ones to stay consistent with Android. Should be good to push, just needs to actually show up on Firebase.

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -277,6 +277,8 @@ class BRBLoginViewController: UIViewController {
     }
     
     @objc private func loginButtonPressed(_ sender: UIButton) {
+        let payload = BRBLoginPressPayload()
+        AppDevAnalytics.shared.log(payload)
```
whoiswillma: I would say to do this as `AppDevAnalytics.shared.log(BRBLoginPressPayload())`

Eatery/Controllers/Eateries/CampusEateriesViewController.swift
```diff
@@ -77,6 +77,9 @@ class CampusEateriesViewController: EateriesViewController {
 
     private func showMenu(of eatery: CampusEatery, animated: Bool) {
         let menuViewController = CampusEateryMenuViewController(eatery: eatery, delegate: self, userLocation: userLocation)
+        let payload: Payload!
```
whoiswillma: No need to have `Payload` as an implicitly unwrapped optional! Just use the ternary directly like this: 

`let payload = eatery.eateryType == .dining ? CampusDiningCellPressPayload() : CollegetownCellPressPayload`

I'm not sure if `eateryType` includes `.dining` but I'm almost certain there is an enum case that matches what you're looking for. Comparing `rawValue` to a string is rather fragile. We don't want to change the `rawValue` and have our analytics break

Eatery/Controllers/Eateries/PillViewController.swift
```diff
@@ -116,11 +116,15 @@ class PillViewController: UIViewController {
     }
 
     @objc private func pillSelectionDidChange() {
+        let payload : Payload!
```
whoiswillma: In this case, Swift is smart enough to determine when the variable is initialized. You can remove the implicity unwrapped optional. 

`let payload: Payload`

And initialize it in each branch of the if

Eatery/Controllers/EateryTabBarController.swift
```diff
@@ -59,4 +59,22 @@ extension EateryTabBarController: UITabBarControllerDelegate {
         return true
     }
 
+    override func tabBar(_ tabBar: UITabBar, didSelect item: UITabBarItem) {
+        let payload : Payload!
```
whoiswillma: Get rid of this variable and call `AppDevAnalytics.shared.log` with the payload initialized within the method call

Eatery/Views/Eateries/FilterBar.swift
```diff
@@ -145,10 +145,41 @@ class FilterBar: UIView {
...
         sender.isSelected.toggle()
-
+        
         if sender.isSelected {
             let filter = displayedFilters[sender.tag]
             selectedFilters.insert(filter)
+
+            // check if filters match that of campus or ctown (taking into account
+            // that nearest first is shared by both)
+            let isCTown = displayedFilters.allSatisfy({ displayedFilter -> Bool in
```
whoiswillma: Instead of having this be done in the filter bar (where we don't necessarily know whether the filters are for Ctown or campus), we can move this into the EateriesVC subclasses.

whoiswillma (changes requested):

Eatery/Views/Eateries/FilterBar.swift
```diff
@@ -145,10 +145,41 @@ class FilterBar: UIView {
...
+                Filter.categoryFilters.contains(displayedFilter) || displayedFilter == Filter.nearest
+            })
+            let payload : Payload!
+            switch filter{
+            case .nearest:
+                payload = isCTown ? CollegetownFilterPressPayload() : NearestFilterPressPayload()
```
whoiswillma: Is `CollegetownFilterPressPayload` the event when the "nearest" filter is pressed when on the CollegetownEateriesVC? If so this name is misleading.

whoiswillma (changes requested):

Eatery/Controllers/BRB/BRBLoginViewController.swift
whoiswillma: There are a bunch of cases where this happens elsewhere. I trust you to find them.

Eatery/Controllers/Eateries/CampusEateriesViewController.swift
```diff
@@ -79,6 +79,24 @@ class CampusEateriesViewController: EateriesViewController {
...
         navigationController?.pushViewController(menuViewController, animated: animated)
+
+        let payload: Payload = eatery.eateryType == .dining ? CampusDiningCellPressPayload() : CampusCafeCellPressPayload()
+        AppDevAnalytics.shared.logFirebase(payload)
+    }
+
+    override func filterBar(_ filterBar: FilterBar, filterWasSelected filter: Filter) {
+        switch filter {
+        case .nearest:
+            AppDevAnalytics.shared.logFirebase(NearestFilterPressPayload())
```
whoiswillma: Line spacing

Eatery/Controllers/Eateries/CollegetownEateriesViewController.swift
```diff
@@ -50,9 +50,14 @@ class CollegetownEateriesViewController: EateriesViewController {
 
     private func showMenu(of eatery: CollegetownEatery) {
         let menuViewController = CollegetownEateriesMenuViewController(eatery: eatery, delegate: self)
+        AppDevAnalytics.shared.logFirebase(CollegetownCellPressPayload())
```
whoiswillma: Move this after line 54

whoiswillma (approved):
LGTM

Eatery/Analytics.swift
```diff
@@ -0,0 +1,136 @@
...
+//  Analytics.swift
+//  AppDevAnalytics
+//
+//  Created by Kevin Chan on 9/4/19.
+//  Copyright © 2019 Kevin Chan. All rights reserved.
+//
+
+import Crashlytics
+import FirebaseAnalytics
+class AppDevAnalytics {
```
kevinchan159, line 12: newline after line 10

Eatery/Analytics.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    static let shared = AppDevAnalytics()
+
+    private init() {}
+
+    func logFabric(_ payload: Payload) {
```
kevinchan159: do we use `logFabric` still? if not, lets remove

Eatery/EventPayload.swift
```diff
@@ -0,0 +1,146 @@
...
+            return ("", nil)
+        }
+
+    }
+
+    func convertToFirebase() -> String {
+
+        let event = self.toEvent()
+
+
```
kevinchan159: delete all these newlines

Eatery/Views/Eateries/FilterBar.swift
```diff
@@ -145,10 +147,12 @@ class FilterBar: UIView {
 
     @objc func buttonPressed(sender: UIButton) {
         sender.isSelected.toggle()
-
+
```
kevinchan159: delete whitespace

Podfile
```diff
@@ -10,7 +10,7 @@ target 'Eatery' do
 
     pod 'ARCL'
     pod 'Apollo', :git => 'https://github.com/apollographql/apollo-ios.git'
-    pod 'AppDevAnalytics', :git => 'https://github.com/cuappdev/ios-analytics.git'
+    #pod 'AppDevAnalytics', :git => 'https://github.com/cuappdev/ios-analytics.git'
```
kevinchan159: delete

kevinchan159 (approved):

Issue #216, opened 2019-09-18
Refactor Eatery Menu View Controller


whoiswillma:
This is a super issue for: 

#203 
#211 
#212 

Goals:
1. Add a common super class for Collegetown and Campus Menu View Controllers
   * Add a common framework for hero animations #211 
   * Remove the navigation bar header for campus eateries #203
2. Re-implement `TabbedPageViewController`

PR #218, opened 2019-09-21
Add navigation bar and title to parallax view controller

whoiswillma:
Issue: #216 

Adds a bunch of enhancements to the parallax view controller: 
* Back button
* Title Label + Gradient
* Accessory View - This can be customized to whatever we want. I wanted to make the `ImageParallaxScrollViewController` not specific to Eatery. In both `MenuViewController`s, ~this will be replaced by the payment view.~ we will add the payment view to the accessory view. For the GIF below I have added a `UIButton` to demonstrate this.

After this PR, I plan to convert both Menu View Controllers to use `ImageParallaxScrollViewController`.

out

### DIFF
--- a/Eatery/Controllers/EateriesNavigationController.swift
+++ b/Eatery/Controllers/EateriesNavigationController.swift
@@ -1,0 +1,30 @@
+//
+//  EateriesNavigationController.swift
+//  ScrollImageView
+//
+//  Created by William Ma on 9/21/19.
+//  Copyright © 2019 William Ma. All rights reserved.
+//
+
+import UIKit
+
+class EateriesNavigationController: UINavigationController {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        delegate = self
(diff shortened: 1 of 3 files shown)


Eatery/Controllers/EateriesNavigationController.swift
```diff
@@ -0,0 +1,33 @@
...
+}
+
+extension EateriesNavigationController: UINavigationControllerDelegate {
+
+    func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
+        if viewController is ImageParallaxScrollViewController {
+            setNavigationBarHidden(true, animated: true)
+        } else {
+            setNavigationBarHidden(false, animated: true)
+        }
```
kevinchan159:
```
let isParallax = viewController is ImageParallaxScrollViewController
setNavigationBarHidden(isParallax, animated: true)
```

kevinchan159 (approved):
very cool!

PR #219, opened 2019-09-21
Change Deployment Target to iOS 11

whoiswillma:
Bump the deployment target to iOS 11. There is little reason to support versions of iOS released over 2 years ago. By updating the deployment target we can use safe area insets and many conveniences of later SDKs.


kevinchan159 (approved):
🚀

chalo2000 (approved):
LGTM!

PR #222, opened 2019-09-28
Tabbed Page View Controller Refactor

whoiswillma:
Issue: #216 

Some issues with the previous implementation of `TabbedPageViewController` (`TabbedPVC`)
1. It was closely coupled with `CampusEateryMealTVC` when it didn't need to be.
2. The `TabbedPVC` had two delegates which made the flow of information confusing. The "regular" delegate was used to update the `UnderlineTabBarView`. The "scroll" delegate was used to notify the `CampusEateryMenuVC` that the meal view controller had changed, and therefore the height of its content needed to be updated.
3. When the `CampusEateryMenuVC` was notified that the meal view controller had changed, it set constraints on the view of the `TabbedPVC`, which meant that additional constraints were being added to the `TabbedPVC` that it was not aware of.
4. The `TabbedPVC` would crash if it was given no view controllers. Although this is guarded against in the `CampusEateryMenuVC`, it would be better if the `TabbedPVC` could ensure itself that it could never access an out of bounds index.

How these issues were solved: 
1. Instead of accessing properties on the `CampusEateryMealTVC`, delegate that functionality out.
2. 
   * I removed the delegate used to update the `UnderlineTabBarView` and instead call methods on `UnderlineTabBarView` directly.
   * The purpose of the scroll delegate was to change the height of the page view controller. Have the page view controller ask its delegate for the height of a specific child and set constraints on itself.
3. Solving (2) also solved (3)
4. Add additional bounds checks. 

Some additional improvements: 
* The `CampusEateryMealTVC` had several implicitly unwrapped optionals that were set upon initialization. Instead, I used an initializer to remove the need for IUOs.

### DIFF
--- a/Eatery/Controllers/Eateries/CampusEateryMealTableViewController.swift
+++ b/Eatery/Controllers/Eateries/CampusEateryMealTableViewController.swift
@@ -2,20 +2,21 @@ import UIKit
 
 class CampusEateryMealTableViewController: UITableViewController {
 
-    var meal: String!
+    let meal: String
+    let eatery: CampusEatery
+    let date: Date
 
-    var eatery: CampusEatery! {
-        didSet {
-            recomputeMenu()
-        }
-    }
-    var event: Event? {
-        didSet {
-            recomputeMenu()
-        }
-    }
+    let event: Event?
 
-    private func recomputeMenu() {
+    private let menu: Menu?
+
+    init(eatery: CampusEatery, meal: String, date: Date) {
+        self.eatery = eatery
+        self.meal = meal
+        self.date = date
+
+        self.event = eatery.eventsByName(onDayOf: date)[meal]
+
         if let eventMenu = event?.menu, !eventMenu.data.isEmpty {
             menu = eventMenu
         } else if eatery.diningMenu != nil, eatery.eateryType != .dining {
@@ -25,21 +26,13 @@ class CampusEateryMealTableViewController: UITableViewController {
             // don't know the menu
             menu = nil
         }
+
+        super.init(nibName: nil, bundle: nil)
     }
 
-    private var menu: Menu? {
-        didSet {
-            if let menu = menu, menu.data.count == 1, eatery.eateryType != .dining {
-                topSeparator.isHidden = false
-                tableView.separatorStyle = .singleLine
-            } else {
-                topSeparator.isHidden = true
-                tableView.separatorStyle = .none
-            }
-        }
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
     }
-
-    private let topSeparator = UIView()
 
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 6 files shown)


Eatery/Controllers/Eateries/CampusEateryMealTableViewController.swift
```diff
@@ -25,21 +26,13 @@ class CampusEateryMealTableViewController: UITableViewController {
             // don't know the menu
             menu = nil
```
kevinchan159, line 27: don't think we need this b/c `menu` is default `nil` right?

Eatery/Controllers/Eateries/TabbedPageViewController.swift
```diff
@@ -1,145 +1,183 @@
...
+//  TabbedPageViewController.swift
+//  Eatery
+//
+//  Created by William Ma on 9/25/19.
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
 import UIKit
 
+// MARK: -
```
kevinchan159: delete or add description?

Eatery/Controllers/Eateries/TabbedPageViewController.swift
```diff
@@ -1,145 +1,183 @@
...
+    func tabbedPageViewController(_ tabbedPageViewController: TabbedPageViewController,
+                                  titleForViewController viewController: UIViewController) -> String?
+
+    func tabbedPageViewController(_ tabbedPageViewController: TabbedPageViewController,
+                                  heightOfContentForViewController viewController: UIViewController) -> CGFloat
+
 }
 
-private let kTabBarHeight: CGFloat = 32.0
+// MARK: -
```
kevinchan159: same here

Eatery/Controllers/Eateries/TabbedPageViewController.swift
```diff
@@ -1,145 +1,183 @@
...
-    
-    var pageViewController: UIPageViewController!
+class TabbedPageViewController: UIViewController {
+
+    let viewControllers: [UIViewController]
+
+    private var pageViewController: UIPageViewController!
+    private var tabBar: UnderlineTabBarView?
+
+    weak var delegate: TabbedPageViewControllerDelegate?
```
kevinchan159, line 28: can we make `private` if possible?

kevinchan159 (approved):
🥳

Eatery/Controllers/Eateries/CampusEateryMealTableViewController.swift
whoiswillma: Swift will complain that `menu` is uninitialized otherwise

Eatery/Controllers/Eateries/TabbedPageViewController.swift
whoiswillma: Referencing our previous conversation Apple's frameworks follow this style of delegate and also this pattern is used throughout all of Eatery currently.

PR #224, opened 2019-10-06
Integrate wait times

chalo2000:
Just creating a PR to have this checked off on - still has bugs.

Basically merged master into ethan/wait_times and got the bars to actually display the backend data (dateFormatter issue was preventing it).

Screen Shot 2019-10-06 at 12 50 22 AM

Current bugs on histogram:
- 10-13m … (make sure string always shows)
- 3am off screen
- Pressing bars near edge of screen cuts off
- Bars aren’t proportional to the numbers showing for wait time

Current bugs caused by merge:
- McCormick's upon press has it's name cut halfway
- Selecting search bar causes the app to freeze


Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -14,6 +14,10 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
 
     var eatery: CampusEatery
     var outerScrollView: UIScrollView!
+    var popularTimesExpanded = true
```
whoiswillma, line 17: Hmmn, @usubzero and I disagreed over whether `PopularTimesXXX` should be a `UIView` or a `UIViewController`. Could @kevinchan159 weigh in on this? I am of the perspective that `PopularTimesView` is the correct level of abstraction, and will continue with my PR review that way.

Edit: Perhaps something Ethan was not aware of when we were writing the histogram originally was that we would be refactoring the Menu View Controllers. Having them as a view would make my job easier when they are refactored.

whoiswillma, line 17: *`PopularTimesView`*

Instead of having a container view, use a `PopularTimesView` which contains `popularTimesToggleButton`, `popularTimesContainer`, and `popularTimesHistogram` refactored to be a `UIView`.

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -177,7 +181,55 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
...
+        popularTimesToggleButton.titleLabel!.font = UIFont.systemFont(ofSize: 14, weight: .medium)
+        popularTimesToggleButton.addTarget(self, action: #selector(popularTimesToggleButtonPressed(sender:)), for: .touchUpInside)
+        
+        popularTimesContainer.addSubview(popularTimesToggleButton)
+        popularTimesToggleButton.snp.makeConstraints { make in
+            make.centerY.equalTo(popularTimesLabel)
+            make.trailing.equalToSuperview().inset(16)
+        }
+        
+        let histogramFrame = CGRect(x: 30, y: 30, width: 350, height: 150)
```
whoiswillma, line 212: Is there any way we can not hard code this frame? Perhaps using the `UIView.layoutSizeFitting` methods?

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -177,7 +181,55 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
             
             return separatorView
         }
-        let infoSeparatorView = makeSeparatorView(topItem: infoContainer, leftInset: 10.0, rightInset: 10.0, topInset: 10.0)
+        let infoSeparatorView1 = makeSeparatorView(topItem: infoContainer, leftInset: 10.0, rightInset: 10.0, topInset: 10.0)
```
whoiswillma: We should find more descriptive names for `infoSeparatorView1` and `infoSeparatorView2` a little farther down

Eatery/Models/Eatery.swift
```diff
@@ -56,6 +56,17 @@ enum EateryStatus {
 
 }
 
+struct SwipeDataPoint {
```
whoiswillma, line 59: This should be moved to `CampusEatery.swift` since `SwipeData` is not something that relates to both Collegetown and Campus eateries.

Eatery/Models/EaterySwipeData.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import Foundation
+
+class EaterySwipeData {
+    
+    private var swipeDataByHours = [Int : [SwipeDataPoint]]()
+  
+    init(swipeDataPoints: [SwipeDataPoint]) {
+        mapSwipeDataPointToHours(swipeDataPoints: swipeDataPoints)
```
whoiswillma, line 16: For type safety, I think three things should happen: 

1. `private let swipeDataByHours: [Int: [SwipeDataPoint]]`
2. `mapSwipeDataPointToHours` functionality should be moved here, either by creating a temporary mutable `var swipeDataByHours` and later assigning it to `self.swipeDataByHours`, or by (3) below. 
3. Use functional methods (probably `reduce(into:)`) to replicate this functionality in one line.

Eatery/Models/EaterySwipeData.swift
```diff
@@ -0,0 +1,58 @@
...
+        for swipeDataPoint in swipeDataPoints {
+            if swipeDataForHour(hour: swipeDataPoint.militaryHour) == nil {
+                swipeDataByHours[swipeDataPoint.militaryHour] = [swipeDataPoint]
+            } else {
+                swipeDataByHours[swipeDataPoint.militaryHour]!.append(swipeDataPoint)
+            }
+        }
+    }
+    
+    func swipeDataForHour(hour: Int) -> [SwipeDataPoint]? {
```
whoiswillma, line 29: If `swipeDataByHours` is immutable, we can remove this method and make `swipeDataByHours` internal.

Eatery/Models/EaterySwipeData.swift
```diff
@@ -0,0 +1,58 @@
...
+        return swipeDataByHours[hour]
+    }
+    
+    func waitTimeFor(hour: Int, minute: Int) -> (waitTimeLow: Int, waitTimeHigh: Int)? {
```
whoiswillma, line 33: We can rename this:
`waitTime(forHour hour: Int, minute: Int) -> (low: Int, high: Int)`

There's no need to namespace the return tuple twice especially when the method name already has `waitTime` in its name. Moreover, it is swift style to include the preposition inside the method parameters.

Eatery/Models/EaterySwipeData.swift
```diff
@@ -0,0 +1,58 @@
...
+        let swipeDataPointsForHour = swipeDataForHour(hour: hour)
+        guard swipeDataPointsForHour != nil else { return nil }
```
whoiswillma: `guard let = swipeDataForHour(hour: hour)` throughout this file.

Eatery/Network/NetworkManager.swift
```diff
@@ -111,6 +111,29 @@ struct NetworkManager {
...
                 }
+                
+                var swipeDataPoints = [SwipeDataPoint]()
+                for swipeDatum in eatery.swipeData {
+                    guard let swipeDatum = swipeDatum else { continue }
+                    
+                    let swipeDatumStartDate = self.timeDateFormatter.date(from: String(swipeDatum.startTime))
+
+                    let swipeDatumEndDate = self.timeDateFormatter.date(from: swipeDatum.endTime)
+                    guard swipeDatumStartDate != nil && swipeDatumEndDate != nil else { continue }
```
whoiswillma, line 122: `guard let`

Eatery/Network/NetworkManager.swift
```diff
@@ -111,6 +111,29 @@ struct NetworkManager {
...
+                    let swipeDatumStartHour = Calendar.current.component(.hour, from: swipeDatumStartDate!)
+                    let swipeDatumStartMinute = Calendar.current.component(.minute, from: swipeDatumStartDate!)
+                    var swipeDatumEndMinute = Calendar.current.component(.minute, from: swipeDatumEndDate!)
+                    swipeDatumEndMinute = (swipeDatumEndMinute == 0) ? 60 : swipeDatumEndMinute
```
whoiswillma, line 126: I feel this invariant should be enforced somewhere other than when we parse it, perhaps inside `SwipeDataPoint`. I had always assumed that `minute` was between 0 and 59.

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import UIKit
+
+// TODO: Create a protocol that encapsulates all Swipe Data and make HistogramViewController independent of Eatery variables
+class HistogramViewController: UIViewController {
+    
+    // MARK: Data
+    var swipeData: EaterySwipeData!
```
whoiswillma, line 15: `private let swipeData: EaterySwipeData`. 

We're passed it in the initializer. 

Edit: And in general, we should find ways to make as many of the non-view properties below constants and private.

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+    }
+    
+    private func setUpAxis() {
+        setUpAxisLineView()
+        
+        let overflowingEndTime = (hourRange.endTime < hourRange.startTime) ? hourRange.endTime + 24 : hourRange.endTime
+        let tickCount: Int = ((overflowingEndTime - hourRange.startTime) / 3) + 1
+        
+        var x: Double = 0
+        for tick in 0..<tickCount {
```
whoiswillma, line 75: Can we find some way to do this more declaratively using constraints instead of merely using constraints as a way to set frames?

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+    
+    // Overflowed military hour is a time between 0...47
+    private func containOverflowedMilitaryHour(hour: Int) -> Int {
+        return (hour >= 24) ? hour - 24 : hour
+    }
+    
+    private func formattedHourForTime(militaryHour: Int) -> String {
+        var formattedHour: String!
```
whoiswillma, line 181:
```
private func string(for militaryHour: Int) -> String {
     let formattedHour: String
}
```

This follows the `NSDateFormatter` method name and style more closely.

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+    }
+    
+    private func setUpBarViews() {
+        setUpBarContainerView()
+        
+        let barCount = (overflowingEndTime - hourRange.startTime) - 2
+        let frameWidth = Double(view.frame.width) - 15
+        let spacingLoss = Double(2 * (barCount - 1))
+        let maxBarHeight = Double(view.frame.height - 102) // 102 is the net space needed for other histogram components
+        barWidth = (frameWidth - spacingLoss) / Double((barCount + 2))
```
whoiswillma, line 127: More math and constants we have to maintain... We should aim to use constraints to remove these constants as well.

whoiswillma (changes requested):
For Gonzalo, the developer in AppDev: Isn't this an odd numbered week? We still have time to polish this up before merging it into master. You still have time to submit another PR within the week to get your 2 PR's in. 

For Gonzalo, the person submitting a PR: I think we should wait on Kevin's opinion as arbitrator about whether to keep wait times as a view controller or view. For the rest of it, to be quite honest, I think most of this code is not optimal and should receive a solid rewrite before its merged into master.

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
kevinchan159: I agree it should be a `UIView`

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -285,7 +337,9 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
         statusLabel.hero.modifiers = fadeModifiers
         // locationImageView.hero.modifiers = fadeModifiers
```
kevinchan159, line 338: can we delete if unused?

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -345,6 +399,28 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
...
     
+    @objc func popularTimesToggleButtonPressed(sender: UIButton) {
+        let newHeight: Int!
+        if popularTimesExpanded {
+            newHeight = 34
+            popularTimesHistogram.view.isHidden.toggle();
+            popularTimesToggleButton.setTitle("Show", for: .normal)
+        } else {
+            newHeight = 180
+            popularTimesHistogram.view.isHidden.toggle()
```
kevinchan159, line 410: this is the same as line 406 so we can move it out of the if else right?

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -345,6 +399,28 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
         mapItem.openInMaps(launchOptions: [MKLaunchOptionsDirectionsModeKey: MKLaunchOptionsDirectionsModeDriving])
     }
     
+    @objc func popularTimesToggleButtonPressed(sender: UIButton) {
+        let newHeight: Int!
```
kevinchan159, line 403: maybe just make this a var with an initial value of 0

Eatery/Models/EaterySwipeData.swift
```diff
@@ -0,0 +1,58 @@
...
+        mapSwipeDataPointToHours(swipeDataPoints: swipeDataPoints)
+    }
+    
+    private func mapSwipeDataPointToHours(swipeDataPoints: [SwipeDataPoint]) {
+        for swipeDataPoint in swipeDataPoints {
+            if swipeDataForHour(hour: swipeDataPoint.militaryHour) == nil {
+                swipeDataByHours[swipeDataPoint.militaryHour] = [swipeDataPoint]
+            } else {
+                swipeDataByHours[swipeDataPoint.militaryHour]!.append(swipeDataPoint)
+            }
```
kevinchan159, line 25: instead of using `!` in line 24, I think we can rewrite it like so:
```
if var dataPoints = swipeDataForHour(hour: swipeDataPoint.militaryHour) {
  dataPoints.append(swipeDataPoint)
} else {
  swipeDataByHours[swipeDataPoint.militaryHour] = [swipeDataPoint]
}
```

Eatery/Models/EaterySwipeData.swift
kevinchan159: can we rename to `getSwipeDataForHour`

Eatery/Models/EaterySwipeData.swift
kevinchan159: might also be worth making `(waitTimeLow: Int, waitTimeHigh: Int)` into a `typealias` of some named type

Eatery/Models/EaterySwipeData.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    func averageSwipeDensityForHour(militaryHour: Int) -> Double {
+        let swipeDataPointsForHour = swipeDataForHour(hour: militaryHour)
+        guard swipeDataPointsForHour != nil else { return 0 }
+        
+        var averageSwipeDensity: Double = 0
+        for swipeDataPoint in swipeDataPointsForHour! {
+            averageSwipeDensity += swipeDataPoint.swipeDensity
+        }
```
kevinchan159, line 44: can simplify this into just a `reduce` function

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+    
+    // MARK: Data
+    var swipeData: EaterySwipeData!
+    // hours in overflowable military time units
+    let hourRange = (startTime: 6, endTime: 3)
+    var overflowingEndTime: Int!
+    
+    // MARK: Views and view properties
+    var viewFrame: CGRect!
+    var axisView: UIView!
```
kevinchan159, line 22: if these view's don't get initialized multiple times, we should make it into something like `private let axisView = UIView()`

Eatery/Views/Eateries/HistogramViewController.swift
kevinchan159: not too sure about the intended design but could we use a `collectionView` here?

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+        let date = Date()
+        let currentHour = Calendar.current.component(.hour, from: date)
+        var waitTimes: (waitTimeLow: Int, waitTimeHigh: Int)!
+        if currentHour == barHour {
+            let currentMinute = Calendar.current.component(.minute, from: date)
+            waitTimes = swipeData.waitTimeFor(hour: currentHour, minute: currentMinute)
+        } else {
+            waitTimes = swipeData.waitTimeFor(hour: barHour, minute: 0)
+        }
+        if waitTimes == nil {
```
kevinchan159, line 246: we shouldn't declare `waitTimes` as `!` if it can actually be `nil`

Eatery/Controllers/Eateries/CampusEateryMenuViewController.swift
```diff
@@ -177,7 +181,55 @@ class CampusEateryMenuViewController: UIViewController, UIScrollViewDelegate, Me
...
+        popularTimesToggleButton.addTarget(self, action: #selector(popularTimesToggleButtonPressed(sender:)), for: .touchUpInside)
+        
+        popularTimesContainer.addSubview(popularTimesToggleButton)
+        popularTimesToggleButton.snp.makeConstraints { make in
+            make.centerY.equalTo(popularTimesLabel)
+            make.trailing.equalToSuperview().inset(16)
+        }
+        
+        let histogramFrame = CGRect(x: 30, y: 30, width: 350, height: 150)
+        //popularTimesHistogram = HistogramViewController(frame: histogramFrame, data: [(2, 4), (4, 6), (3, 7), (5, 7), (1, 3), (7, 9), (2, 4), (4, 6), (3, 7), (5, 7), (1, 3), (7, 9), (2, 4), (4, 6), (3, 7), (5, 7), (1, 3), (7, 9), (2, 4), (4, 6), (3, 7)])
```
young-k, line 213: delete if unused?

Eatery/Extensions/UIColor+ColorScheme.swift
```diff
@@ -26,6 +26,9 @@ extension UIColor {
 
     /// 0x4A90E2
     static let eateryBlue = colorFromCode(0x4A90E2)
+    
+    /// 0xb7d3f3
```
young-k, line 30: What's the point of having the hex in a comment here (and in the rest of this file)?

Eatery/Views/Eateries/HistogramViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+            tickLabel.snp.makeConstraints { make in
+                make.bottom.equalTo(view).inset(10)
+                make.leading.equalTo(view).offset(x)
+                make.width.equalTo(30)
+                make.height.equalTo(15)
+            }
+            
+            let additionalSpacing = !(tick == 0 || tick >= (tickCount - 2))
+            let xSpacingAdd: Double = (((additionalSpacing ? 2 : 0) + 1) * 2) + ((tick >= (tickCount - 2)) ? 4 : 0)
+            x += (3 * barWidth) + xSpacingAdd
```
young-k, line 104: these two lines are pretty difficult to read, could we add comments here?

whoiswillma (approved):

PR #230, opened 2019-10-12
Wait Time Improvements

whoiswillma:
- Add "No Estimate" text to wait times
- Hide popular times by default if there is no wait time data
- Allow the user to drag their finger below the bounds of the view when swiping between swipe data
- Haptics


young-k (approved):
ship it

kevinchan159 (approved):
lets gooooo

PR #235, opened 2019-10-16
Change amount color in BRBHistoryViewController

ReadeP:
## Summary
Originally there was no way to differentiate between positive transactions (money enters user's account) and negative transactions (money leaves user's account). These changes show incoming money as green and positive.

## Changes Made
- Updated brbQuery.graphql by adding positive attribute to history
- Added 'positive' boolean attribute to BRBAccount model
- Updated brbHistory array in NetworkManager to return BRBHistory object with a value for the positive boolean
- Updated the configuration of BRBHistoryTableViewCell to check the positive value and if positive, display the amount in green and if negative, display it in red

## Test Coverage
I tested these changes initially by hardcoding the positive values for each cell based on a random boolean against my own account since I did not received any money in my account. I also tested it against William's account which had positive transactions.

## Screenshots
Simulator Screen Shot - iPhone 11 - 2019-10-16 at 18 12 10
Simulator Screen Shot - iPhone 11 - 2019-10-16 at 18 12 06
Simulator Screen Shot - iPhone 11 - 2019-10-16 at 18 12 08


Eatery/Views/BRB/BRBHistoryTableViewCell.swift
```diff
@@ -13,6 +13,7 @@ class BRBHistoryTableViewCell: UITableViewCell {
     private var titleLabel: UILabel!
     private var subtitleLabel: UILabel!
     private var priceLabel: UILabel!
+    private var positive: Bool!
```
whoiswillma: We don't need this property; the label text color can be set inside the `configure` method.

Eatery/Views/BRB/BRBHistoryTableViewCell.swift
```diff
@@ -53,10 +53,18 @@ class BRBHistoryTableViewCell: UITableViewCell {
         fatalError("init(coder:) has not been implemented")
     }
 
-    func configure(title: String, subtitle: String, amount: String) {
-        titleLabel.text = title
-        subtitleLabel.text = subtitle
-        priceLabel.text = "-$\(amount)"
+    func configure(title: String, subtitle: String, amount: String, positive: Bool) {
+        self.titleLabel.text = title
```
whoiswillma: Can we get rid of the `self.`?

whoiswillma (approved):
Looks good!

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -86,7 +86,7 @@ extension BRBAccountViewController: UITableViewDataSource {
         let historyItem = account.history[indexPath.row]
         cell.configure(title: historyItem.name,
                        subtitle: historyItem.timestamp,
-                       amount: historyItem.amount)
+                       amount: historyItem.amount, positive: historyItem.positive)
```
kevinchan159: can we move `positive: historyItem.positive` onto its own line like the other arguments? i.e. it should look like
```
cell.configure(
  title: historyItem.name,
  subtitle: historyItem.timestamp,
  amount: historyItem.amount
  positive: historyItem.positive
)
```

kevinchan159 (approved):
amazing!!! 🥳

PR #236, opened 2019-10-16
Update popular times view

whoiswillma:
## Overview

The current time is now shown always on the popular times view

## Changes Made

The current time is now shown always on the popular times view

## Test Coverage

I used the app


kevinchan159 (approved):
lgtm!

PR #238, opened 2019-10-21
Select Relevant Meal on Menu

whoiswillma:
## Overview

Present a meal based on the current time

## Changes Made

I use the active event of the eatery to determine which meal is currently active, then change the index of the `TabbedPageViewController` to reflect the active meal.

## Test Coverage

I waited until noon today to see if lunch would be selected.

### DIFF
--- a/Eatery/Controllers/Eateries/CampusMenuViewController.swift
+++ b/Eatery/Controllers/Eateries/CampusMenuViewController.swift
@@ -204,6 +204,16 @@ class CampusMenuViewController: ImageParallaxScrollViewController {
         pageViewController.didMove(toParentViewController: self)
 
         pageViewController.view.hero.modifiers = createHeroModifiers(.fade, .translate)
+
+        if let currentEvent = eatery.activeEvent(atExactly: Date()) {
+            for (index, viewController) in viewControllers.enumerated() {
+                if currentEvent.desc == "Lite Lunch", viewController.meal == "Lunch" {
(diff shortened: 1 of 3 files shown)


Eatery/Controllers/Eateries/CampusMenuViewController.swift
```diff
@@ -204,6 +204,16 @@ class CampusMenuViewController: ImageParallaxScrollViewController {
         pageViewController.didMove(toParentViewController: self)
 
         pageViewController.view.hero.modifiers = createHeroModifiers(.fade, .translate)
+
+        if let currentEvent = eatery.activeEvent(atExactly: Date()) {
+            for (index, viewController) in viewControllers.enumerated() {
+                if currentEvent.desc == "Lite Lunch", viewController.meal == "Lunch" {
+                    pageViewController.currentViewControllerIndex = index
+                } else if viewController.event?.desc == currentEvent.desc {
+                    pageViewController.currentViewControllerIndex = index
```
ReadeP, line 212: Why is this an else if statement if it does the same thing?

ReadeP (approved):
🤩

Eatery/Controllers/Eateries/CampusMenuViewController.swift
whoiswillma: Thank you for pointing this out! I'll fix it

PR #241, opened 2019-10-25
Remove Crashalytics, iOS 11 and Below checks

whoiswillma:
## Overview

Crashalytics has been deintegrated. I made some other small changes as well
* Remove `MenuImages.swift` which has been sitting unused for some time
* Remove `if #available` checks for iOS 11 and below since the minimum version we support is iOS 11.


kevinchan159 (approved):
lgtm!

ReadeP (approved):
👏

PR #245, opened 2019-11-02
Query BRB account data on launch

ReadeP:
##  Overview

Prior, when users wanted to check their BRB account, they would have to login each time they started the app. In this PR, I automated the process of login by having automatically requesting the user's account data from using their saved credentials on launch. While this is happening, if the user views their account, they will see their cached data with an activity icon in the navigation bar.

## Changes Made

- Refactored `BRBAccountSettings` and `BRBConnectionHandler` into one file called `BRBAccountManager` when the app launches
- Automatically fetches the user's login credentials from keychain and makes a request in `EateryTabBarController`
- Used UserDefaults to cache user's BRBAccount data
- Changes to `BRBViewController` navigation bar, including an activity indicator in the top left corner to show the status of the request and changed the title to say 'Hello, NetID' when the user is logged in
- Added a refresh control to tableview in `AccountViewController` to trigger a new request
- Removed the ability to save login credentials on `BRBLoginViewController`. Instead, Eatery automatically saves a user's credentials until they log out.

## Test Coverage

I manually tested this feature by launching the app with and without having previously logged in. I used both real data and mock data to test if the `BRBAccountViewController` was being updated appropriately. I also tested its functionality without internet connection to ensure the request was handling errors correctly.

## Next Steps

Currently, when the request fails, a `UIAlertController` is presented. In the future, I think we could work with design to build a more embedded error view in the `BRBViewController` like the one in `BRBLoginViewController` when a login fails. Additional next steps are to start working on on-boarding.

## Screenshots

<details>

  <summary>BRBAccountViewController While Requesting New Data</summary>

Simulator Screen Shot - iPhone 11 - 2019-11-02 at 13 03 11

</details>

<details>
  <summary>Error Message When Request Fails</summary>

Simulator Screen Shot - iPhone 11 - 2019-11-02 at 12 52 38

</details>


Eatery/AppDelegate.swift
```diff
@@ -13,7 +13,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var eateryTabBarController: EateryTabBarController!
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions:  [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-
+
```
whoiswillma: Remove space

To save you the effort of having to remove it everywhere, enable "automatically trim trailing whitespace" and "including whitespace only lines": 
Screen Shot 2019-11-02 at 13 06 24

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -13,6 +13,7 @@ import SafariServices
 protocol AboutTableViewControllerDelegate: AnyObject {
 
     func aboutTableViewControllerDidLogoutUser(_ stvc: AboutTableViewController)
+    func didClickBackButton()
```
whoiswillma: Can we rename this to be more stylistically like other delegates? Include the caller (`AboutTableViewController`) and also rename `click` to `tap` for iOS.

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -54,6 +55,13 @@ class AboutTableViewController: UITableViewController {
         tableView.register(SwitchTableViewCell.self, forCellReuseIdentifier: CellIdentifier.saveLoginInfo.rawValue)
         tableView.register(LogoutTableViewCell.self, forCellReuseIdentifier: CellIdentifier.logout.rawValue)
     }
+    
+    override func willMove(toParentViewController parent: UIViewController?) {
+        super.willMove(toParentViewController: parent)
+        if parent == nil {
```
whoiswillma, line 62: Can you add a comment or explain why this is needed?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+
+private let loginURLString = "https://get.cbord.com/cornell/full/login.php?mobileapp=1"
+private let maxTrials = 3
+private let trialDelay = 500
+
+//MARK: -
+//MARK: Account Manager
+
+protocol BRBAccountManagerDelegate {
+    func failedToGetAccount(with error: String)
```
whoiswillma: Same thing with these two delegates as well. 

I know these were in the Eatery codebase like this before, but when we touch something we should always try and make it cleaner than we found it.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+
+    func retrievedSessionId(id: String)
+    func loginFailed(with error: String)
+
+}
+
+private class BRBConnectionHandler: WKWebView, WKNavigationDelegate {
+    
+    var stage: Stages = .loginScreen
+    var accountBalance: AccountBalance!
```
whoiswillma: Is this used any where? I just did `CMD+F` and couldn't find any usages. Maybe get rid of it?

whoiswillma: May also be a good idea to make some of these private as well

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+    //MARK: Connection Handling
+    
+    /**
+     - Gets the HTML for the current web page and runs block after loading HTML into a string
+     */
+    func getHTML(block: @escaping (NSString) -> ()){
+        evaluateJavaScript("document.documentElement.outerHTML.toString()",
+                           completionHandler: { (html: Any?, error: Error?) in
+                            if error == nil {
+                                block(html as! NSString)
```
whoiswillma: Convert this to safe unwrap

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+     - Loads login web page
+     */
+    func handleLogin() {
+        loginCount = 0
+        stage = .loginScreen
+
+        // Remove cache
+        URLCache.shared.removeAllCachedResponses()
+
+        let loginURL = URL(string: loginURLString)!
```
whoiswillma: Safe unwrap

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+        let keychain = KeychainItemWrapper(identifier: "netid", accessGroup: nil)
+        guard let netid = keychain["netid"] as? String, let password = keychain["password"] as? String else {
+            return nil
+        }
+        return (netid: netid, password: password)
+    }
+    
+}
+
+private let loginURLString = "https://get.cbord.com/cornell/full/login.php?mobileapp=1"
```
whoiswillma: Perhaps convert this to a `URL` instead of a string that we later parse into a URL?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+            } else {
+                self.loginFailed(with: "Unable to parse account")
+            }
+        }
+    }
+    
+    func loginFailed(with error: String) {
+        self.delegate?.failedToGetAccount(with: error)
+    }
+
```
whoiswillma, line 264: I think with eatery we usually have either zero or one newlines at the end of class definitions 

```
     func { 
     }
}
```

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -110,8 +157,12 @@ extension BRBViewController: BRBLoginViewControllerDelegate {
     
 }
 
-extension BRBViewController: BRBConnectionDelegate {
+extension BRBViewController: BRBAccountManagerDelegate {
     
+    func queriedAccount(account: BRBAccount) {
```
whoiswillma: I tried  pulling your branch and could not get it to compile because this function is incomplete, make sure to check that everything compiles before submitting a PR!

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,70 +16,105 @@ class BRBViewController: UIViewController {
...
+    private var loggedIn = false
     
     private lazy var loginViewController = BRBLoginViewController()
-    private var accountViewController: BRBAccountViewController?
+    var accountViewController: BRBAccountViewController?
     
     private var state: State = .login
+    
+    private var activityIndicator: NVActivityIndicatorView?
+    var isLoading: Bool? {
```
whoiswillma: Can we get rid of `isLoading`? Seems all we use it for is to update the `activityIndicator`.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+    func removeSavedLoginInfo() {
+        BRBAccountSettings.removeKeychainLoginInfo()
+    }
+    
+    func getCredentials() -> LoginInfo? {
+        return BRBAccountSettings.loadFromKeychain()
+    }
+    
+    @objc func queryCachedBRBData() {
+        if let (netid, password) = BRBAccountSettings.loadFromKeychain() {
```
whoiswillma: I think we should rename this: `queryCachedBRBData` sounds like it should notify the delegate with the cached account data. The login data is saved more than it is cached Perhaps `queryBRBDataWithSavedLogin()` might be better.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+            connectionHandler.netid = netid
+            connectionHandler.password = password
+            connectionHandler.handleLogin()
```
whoiswillma: Call `queryBRBData(netid:, password:)` to remove duplicate code!

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,296 @@
...
+    }
+    
+    func queryBRBData(netid: String, password: String) {
+        connectionHandler.netid = netid
+        connectionHandler.password = password
+        connectionHandler.handleLogin()
+    }
+    
+    func getConnectionStage() -> Stages {
+        return connectionHandler.stage
```
whoiswillma: Have this be a get-only property?

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,70 +16,105 @@ class BRBViewController: UIViewController {
...
+            activityIndicator = NVActivityIndicatorView(frame: CGRect(x: 0, y: 0, width: 22, height: 22), type: .circleStrokeSpin, color: .white)
+            let activityItem = UIBarButtonItem(customView: activityIndicator!)
+            navigationItem.setLeftBarButton(activityItem, animated: true)
+            
+            let refreshControl = UIRefreshControl(frame: .zero)
+            refreshControl.tintColor = .white
+            refreshControl.addTarget(self, action: #selector(refreshBRBAccount), for: .valueChanged)
+            self.accountViewController?.tableView.refreshControl = refreshControl
+            if accountManager.getCredentials() != nil {
+                navigationItem.title = "Hello, \(accountManager.getCredentials()!.netid)"
```
whoiswillma: Remove force unwrap

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,70 +16,105 @@ class BRBViewController: UIViewController {
...
             accountViewController.didMove(toParentViewController: self)
             
+            activityIndicator = NVActivityIndicatorView(frame: CGRect(x: 0, y: 0, width: 22, height: 22), type: .circleStrokeSpin, color: .white)
+            let activityItem = UIBarButtonItem(customView: activityIndicator!)
+            navigationItem.setLeftBarButton(activityItem, animated: true)
+            
+            let refreshControl = UIRefreshControl(frame: .zero)
+            refreshControl.tintColor = .white
+            refreshControl.addTarget(self, action: #selector(refreshBRBAccount), for: .valueChanged)
+            self.accountViewController?.tableView.refreshControl = refreshControl
```
whoiswillma: I think the `accountViewController` should manage its own refresh control. Maybe have a property called `isRefreshing` in `accountViewController` that has a `didSet`. A `get set` property might work as well.

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -91,6 +127,19 @@ class BRBViewController: UIViewController {
...
+    func showErrorAlert(error: String) {
+        let errorAlert = UIAlertController(title: error, message: "Unable to fetch new BRB account data. Please check your connection and try again.", preferredStyle: UIAlertControllerStyle.alert)
+        errorAlert.addAction(UIAlertAction(title: "Okay", style: UIAlertActionStyle.destructive, handler: nil))
+        self.present(errorAlert, animated: true, completion: nil)
+        isLoading = false
+    }
+    
+    @objc private func refreshBRBAccount(_ sender: Any) {
+        accountManager.queryCachedBRBData()
+        accountViewController?.tableView.refreshControl?.endRefreshing()
```
whoiswillma: Same thing about the account view controller here

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -129,23 +180,39 @@ extension BRBViewController: BRBConnectionDelegate {
...
 extension BRBViewController: AboutTableViewControllerDelegate {
-    
+
     func aboutTableViewControllerDidLogoutUser(_ stvc: AboutTableViewController) {
         setState(.login)
-        
-        connectionHandler = BRBConnectionHandler()
-        connectionHandler.delegate = self
-        
+        UserDefaults.standard.set(nil, forKey: "BRBAccount")
```
whoiswillma: Clearing of cache should be handled by `BRBAccountManager` b/c how the information is cached should be invisible to the rest of the application

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -129,23 +180,39 @@ extension BRBViewController: BRBConnectionDelegate {
...
+        navigationItem.title = "Meal Plan"
+        accountManager = BRBAccountManager()
+        accountManager.delegate = self
+        accountManager.removeSavedLoginInfo()
```
whoiswillma, line 177: Do we need to recreate the account manager here? If this is used to reset the connection handler, then maybe adding a method on `BRBAccountManager` called `resetConnection` might be better.

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -129,23 +180,39 @@ extension BRBViewController: BRBConnectionDelegate {
...
+        loggedIn = false
+        isLoading = false
         navigationController?.popToViewController(self, animated: true)
     }
     
+    func didClickBackButton() {
+        if accountManager.getCredentials() != nil {
```
whoiswillma: `if let`

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -298,10 +219,8 @@ class BRBLoginViewController: UIViewController {
         netidTextField.resignFirstResponder()
         passwordTextField.resignFirstResponder()
         
-        if BRBAccountSettings.saveLoginInfo {
-            let loginInfo = (netid: netid, password: password)
-            BRBAccountSettings.saveToKeychain(loginInfo: loginInfo)
-        }
+        let loginInfo = (netid: netid, password: password)
+        accountManager?.saveLoginInfo(loginInfo: loginInfo)
```
whoiswillma: We can get rid of the `accountManager` from the `LoginViewController` by just having the delegate of the `LoginViewController` (i.e. the `BRBViewController`) save the login info automatically whenever the user logs in.

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,70 +16,105 @@ class BRBViewController: UIViewController {
         case account(BRBAccount)
     }
     
-    private var connectionHandler: BRBConnectionHandler!
+    var accountManager: BRBAccountManager!
```
whoiswillma: Initialize inline with let to be more typesafe

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,70 +16,105 @@ class BRBViewController: UIViewController {
...
     private var requestStart: Date?
+    private var account: BRBAccount?
+    private var loggedIn = false
     
     private lazy var loginViewController = BRBLoginViewController()
-    private var accountViewController: BRBAccountViewController?
+    var accountViewController: BRBAccountViewController?
```
whoiswillma: Have this be private?

Eatery/Controllers/EateryTabBarController.swift
```diff
@@ -15,10 +15,13 @@ class EateryTabBarController: UITabBarController {
     let eateriesSharedViewController = EateriesSharedViewController()
     let lookAheadViewController = LookAheadViewController()
     let brbViewController = BRBViewController()
+    let accountManager = BRBAccountManager()
```
whoiswillma: Suggestion: Remove this from `EateryTabBarController`; I know I mentioned that you should do this before but I had a different implementation in mind (although yours works as well). 

I think we can make the call to `queryCachedBRBData` (or whatever you decide to rename it to, if you take my suggestion) inside the `BRBViewController`. The idea being that there should only be one `BRBAccountManager` in existence at a time.

whoiswillma (changes requested):
Lots of asks and PR comments, but great job on implementing your first large feature!

Eatery/Controllers/BRB/BRBAccountManager.swift
whoiswillma: Or 

```
    func {
    }

}
```

Eatery/Controllers/BRB/BRBViewController.swift
whoiswillma: I think this may have been caused by a merge conflict. Git will try its best to combine master with your branch, but it may not do a perfect job. Sometimes you'll get merge conflicts, other times it'll do things like this. I usually run through all the changes and try to compile the app before submitting PRs for this reason

ReadeP: Yes, I saw that there was a merge conflict in this function, which I believe I fixed

Eatery/Controllers/BRB/BRBViewController.swift
ReadeP: We access isLoading in EateryTabBarController

Eatery/Controllers/BRB/BRBAccountManager.swift
ReadeP: Good catch!

Eatery/Controllers/BRB/BRBViewController.swift
whoiswillma: At the time I made this comment I hadn't seen `EateryTabBarController`, but with the additional comments I made below I still think we should remove this property.

ReadeP: Wouldn't that mean we would have to unwrap activityIndicator a lot throughout the file? isLoading embeds that check, which is much cleaner I feel

whoiswillma: We can just do `activityIndicator?.startAnimating()`

whoiswillma: Wait, better suggestion. Make activity indicator not an optional. `private let activityIndicator = NVActivityIndicatorView(< blah >)`

Eatery/Controllers/BRB/BRBViewController.swift
ReadeP: Yeah I was thinking this too

Eatery/AppDelegate.swift
Omarrasheed: is swiftlint not setup on repo?

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -13,6 +13,7 @@ import SafariServices
 protocol AboutTableViewControllerDelegate: AnyObject {
 
     func aboutTableViewControllerDidLogoutUser(_ stvc: AboutTableViewController)
+    func aboutTaleViewControllerDidTapBackButton(_ stvc: AboutTableViewController)
```
Omarrasheed: misspelled `table`

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+
+    case loginScreen
+    case transition
+    case loginFailed
+    case finished(sessionId: String)
+
+}
+
+//MARK: -
+//MARK: Account Settings
```
Omarrasheed: Should just be using `// MARK: - Account Settings` cc @kevinchan159

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+    }
+    
+}
+
+private let loginURLString = "https://get.cbord.com/cornell/full/login.php?mobileapp=1"
+private let maxTrials = 3
+private let trialDelay = 500
+
+//MARK: -
+//MARK: Account Manager
```
Omarrasheed: Same here

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+        URLCache.shared.diskCapacity = 0
+        URLCache.shared.memoryCapacity = 0
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    //MARK: -
+    //MARK: Connection Handling
```
Omarrasheed: Might wanna do a sweep of the codebase on this

Eatery/Controllers/BRB/AboutTableViewController.swift
CameronHamidi: Should this be "Table" - typo?

ReadeP: Yeah that's a typo

jackthmp: Typo!

danielVebman: Typo, remove newline.

Eatery/Controllers/BRB/BRBAccountManager.swift
danielVebman: Combine into `// MARK: - Account Manager`

Omarrasheed (commented):
Looks dope. Some small nits for ya

Eatery/Controllers/BRB/BRBAccountManager.swift
danielVebman: Same comment as above. Combine these two.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+    
+    /**
+     - Gets the HTML for the current web page and runs block after loading HTML into a string
+     */
```
danielVebman: Rewrite as `/// Gets the ...`. Google's Swift style guide says no to Java style comments.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+    func getHTML(block: @escaping (NSString) -> ()){
```
danielVebman: NSString? Yuck! Why not a String?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+    }
+    
+    func failedToLogin() -> Bool {
+        return loginCount > 1
+    }
+    
+    @objc func login() {
+        let javascript = "document.getElementsByName('netid')[0].value = '\(netid)';document.getElementsByName('password')[0].value = '\(password)';document.forms[0].submit();"
+        
+        evaluateJavaScript(javascript){ (result: Any?, error: Error?) -> Void in
```
danielVebman: Space after `(javascript)`

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+
+protocol BRBAccountManagerDelegate {
+    func BRBAccountManagerDidFailToQueryAccount(with error: String)
+    func BRBAccountManagerDidQueryAccount(account: BRBAccount)
+}
+
+private protocol BRBConnectionDelegate {
+
+    func BRBConnectionHandlerDelegateDidRetrieveSessionId(id: String)
+    func BRBConnectionHandlerDelegateDidFailLogin(with error: String)
```
danielVebman: Functions should start lowercase, and you can shorten this:

    brbConnectionHandlerDidRetrieve(sessionId: String)

Same thing for `BRBAccountMangerDelegate` methods.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+    }
+    
+    //MARK: -
+    //MARK: Connection Handling
+    
+    /**
+     - Gets the HTML for the current web page and runs block after loading HTML into a string
+     */
+    func getHTML(block: @escaping (NSString) -> ()){
+        evaluateJavaScript("document.documentElement.outerHTML.toString()",
```
danielVebman, line 90: Does fetching BRBs need to include evaluating JavaScript?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+        self.delegate?.BRBConnectionHandlerDelegateDidFailLogin(with: error.localizedDescription)
+    }
+    
+    /**
+     - Gets the stage enum for the currently displayed web page and runs a block after fetching
+     the HTML for the page.
+     
+     - Does not guarantee Javascript will finish running before the block
+     is executed.
+     */
```
kevinchan159: `///` instead of java doc comments

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,287 @@
...
+                if let sessionId = urlComponents.queryItems?.first(where: { $0.name == "sessionId" })?.value {
+                    self.stage = .finished(sessionId: sessionId)
+                } else {
+                    self.stage = .loginFailed
+                }
+            } else {
+                self.stage = .transition
+            }
+            
+            //run block for stage
```
kevinchan159: `// Run block for stage`

Eatery/Controllers/Eateries/EateriesSharedViewController.swift
```diff
@@ -46,7 +46,7 @@ class EateriesSharedViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-
+
```
kevinchan159: remove whitespace

Eatery/Controllers/BRB/BRBAccountManager.swift
ReadeP: @whoiswillma

whoiswillma: Maybe not, but at this point I'm too afraid to touch it. I'm not too familiar with `WKWebView` or web technologies in general

whoiswillma: If you (anyone) knows a better way to do this then I'm all for it.

Eatery/AppDelegate.swift
```diff
@@ -12,7 +12,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
```
kevinchan159: bump on this

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -13,6 +13,7 @@ import SafariServices
 protocol AboutTableViewControllerDelegate: AnyObject {
 
     func aboutTableViewControllerDidLogoutUser(_ stvc: AboutTableViewController)
+    func aboutTableViewControllerDidTapBackButton(_ stvc: AboutTableViewController)
```
kevinchan159: what does `stvc` stand for?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+enum Stages {
+
+    case loginScreen
+    case transition
+    case loginFailed
+    case finished(sessionId: String)
+
+}
+
+//MARK: - Account Settings
```
kevinchan159: add space after `//`

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+private protocol BRBConnectionHandlerDelegate {
+
+    func brbConnectionHandlerDelegateDidRetrieve(sessionID: String)
+    func brbConnectionHandlerDelegateDidFailLogin(with error: String)
+
+}
+
+private class BRBConnectionHandler: WKWebView, WKNavigationDelegate {
+    
+    var stage: Stages = .loginScreen
```
kevinchan159: i think the enum should be renamed to `Stage` because `Stages` seems to imply multiple

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -43,6 +49,17 @@ class BRBAccountViewController: UIViewController {
         tableView.snp.makeConstraints { make in
             make.edges.equalToSuperview()
         }
+
+        refreshControl = UIRefreshControl(frame: .zero)
+        refreshControl.tintColor = .white
+        refreshControl.addTarget(self, action: #selector(refreshBRBAccount), for: .valueChanged)
+        tableView.refreshControl = refreshControl
+
```
kevinchan159: delete newline

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -105,4 +122,10 @@ extension BRBAccountViewController: UITableViewDelegate {
         return nil
     }
     
+    func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+        if section == 0 {
+            return 30
+        }
+        return UITableViewAutomaticDimension
```
kevinchan159: return `section == 0 ? 30 : UITableViewAutomaticDimension`

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -121,12 +80,13 @@ class BRBLoginViewController: UIViewController {
...
         setUpPasswordViews()
-        setUpSaveLoginInfoViews()
         setUpLoginViews()
+    }
 
-        if let (netid, password) = BRBAccountSettings.loadFromKeychain(), !netid.isEmpty, !password.isEmpty {
-            netidTextField.text = netid
-            passwordTextField.text = password
+    override func viewWillAppear(_ animated: Bool) {
+        if netidTextField.text != nil && passwordTextField.text != nil {
```
kevinchan159: use `isEmpty` check instead? i.e. `!netidTextField.text.isEmpty`

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,26 +16,41 @@ class BRBViewController: UIViewController {
...
```
kevinchan159: why do we need to make this `public`?

Eatery/Controllers/EateryTabBarController.swift
```diff
@@ -18,7 +18,7 @@ class EateryTabBarController: UITabBarController {
 
     override func viewDidLoad() {
         delegate = self
-
+
```
kevinchan159: delete whitespace

kevinchan159 (commented):
looks great! 🚀  some minor comments

Eatery/Controllers/BRB/AboutTableViewController.swift
ReadeP: I am not entirely sure, stvc was defined in the logout protocol function, so I followed that format. stvc is never used so I am not sure why we even need it @whoiswillma?

Eatery/Controllers/BRB/BRBViewController.swift
ReadeP: It should be private. In a previous commit it was being accessed from the EateryTabBarController, but no longer is.

Eatery/Controllers/BRB/AboutTableViewController.swift
danielVebman: `stvc` seems to refer to `AccountTableViewController`. However, doing a quick search through the repo on GitHub, it seems no class actually inherits from the delegate, so you could probably cut it.

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -55,6 +56,15 @@ class AboutTableViewController: UITableViewController {
         tableView.register(LogoutTableViewCell.self, forCellReuseIdentifier: CellIdentifier.logout.rawValue)
     }
 
+    // Detects when user taps the 'Back' button in navigation bar
+    // Source: https://stackoverflow.com/a/32245222
+    override func willMove(toParentViewController parent: UIViewController?) {
+        super.willMove(toParentViewController: parent)
+        if parent == nil {
```
danielVebman, line 62: Newline after `super. ...`.

But I think you might be able to cut this because the delegate is always `nil` (*verify that*).

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+        }
+        return (netid: netid, password: password)
+    }
+    
+}
+
+private let loginURL = URL(string: "https://get.cbord.com/cornell/full/login.php?mobileapp=1")
+private let maxTrials = 3
+private let trialDelay = 500
+
```
danielVebman: What are these variables? Why are they global? It seems the latter 2 aren't even used, and `loginURL` is only used within `BRBConnectionHandler`.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+private class BRBConnectionHandler: WKWebView, WKNavigationDelegate {
+    
+    var stage: Stage = .loginScreen
+    private var diningHistory: [HistoryEntry] = []
+    private var loginCount = 0
+    var netid: String = ""
+    var password: String = ""
+    var delegate: BRBConnectionHandlerDelegate?
+
```
danielVebman, line 74: Put these in alphabetical order. Maybe do private vars on top in alphabetical order, then public variables right under also in alphabetical order:

ex. 

    private var a
    private var d

    var b
    var c

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+    
+    var stage: Stage = .loginScreen
+    private var diningHistory: [HistoryEntry] = []
+    private var loginCount = 0
+    var netid: String = ""
+    var password: String = ""
+    var delegate: BRBConnectionHandlerDelegate?
+    
+    init() {
+        super.init(frame: .zero, configuration: WKWebViewConfiguration())
```
danielVebman, line 76: Newline after.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+}
+
+enum Stage {
+
+    case loginScreen
+    case transition
+    case loginFailed
+    case finished(sessionId: String)
+
+}
```
danielVebman, line 23: I would remove the extra newlines for the struct and enum.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+    /// Gets the HTML for the current web page and runs block after loading
+    /// HTML into a string
+    func getHTML(block: @escaping (String) -> ()){
+        evaluateJavaScript("document.documentElement.outerHTML.toString()",
+                           completionHandler: { (html: Any?, error: Error?) in
+                            if let html = html {
+                                block(html as! String)
+                            }
+        })
+    }
```
danielVebman, line 95: Force unwraps are generally bad. Will the compiler take this:

    evaluateJavaScript("document.documentElement.outerHTML.toString()") { (html: Any?, error: Error?) in
        guard let html = html as? String else { return}
        block(html)
    })

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+        connectionHandler.password = password
+        connectionHandler.handleLogin()
+    }
+    
+    func getCachedAccount() -> BRBAccount? {
+        if let brbAccount = UserDefaults.standard.object(forKey: "BRBAccount") as? Data {
+            let decoder = JSONDecoder()
+            if let loadedBRBAccount = try? decoder.decode(BRBAccount.self, from: brbAccount) {
+                return loadedBRBAccount
+            }
```
danielVebman: Replace lines 242-9 with:

```
if let accountData = UserDefaults.standard.object(forKey: "BRBAccount") as? Data,
    let account = try? JSONDecoder().decode(BRBAccount.self, from: accountData) {
    return account
}

return nil
```

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+
+// MARK: - Connection Handler
+
+private protocol BRBConnectionHandlerDelegate {
+
+    func brbConnectionHandlerDelegateDidRetrieve(sessionID: String)
+    func brbConnectionHandlerDelegateDidFailLogin(with error: String)
+
+}
+
```
danielVebman, line 63: These names are *long*. Replace with:

    brbConnectionHandler(didRetrieve sessionID: String)
    brbConnectionHandler(didFailWith error: String)

A rather pedantic note: It is not the *delegate* that did retrieve or did fail. Rather, it is the `brbConnectionHandler`. The method names should reflect that.

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -121,12 +80,13 @@ class BRBLoginViewController: UIViewController {
...
         setUpLoginViews()
+    }
 
-        if let (netid, password) = BRBAccountSettings.loadFromKeychain(), !netid.isEmpty, !password.isEmpty {
-            netidTextField.text = netid
-            passwordTextField.text = password
+    override func viewWillAppear(_ animated: Bool) {
+        if !(netidTextField.text?.isEmpty ?? true) && !(passwordTextField.text?.isEmpty ?? true) {
+            netidTextField.text = nil
+            passwordTextField.text = nil
```
danielVebman: This if statement is kind of complicated. Could you add a short explanation of what the condition is? Even better: could you make it simpler?

Eatery/Controllers/BRB/BRBViewController.swift
danielVebman: ^

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -43,33 +65,46 @@ class BRBViewController: UIViewController {
             make.edges.equalToSuperview()
         }
         loginViewController.didMove(toParentViewController: self)
+        
+//        if let account = accountManager.getCachedAccount() {
+//            setState(.account(account))
+//            activityIndicator.startAnimating()
+//            loggedIn = true
+//        }
```
danielVebman: Can you just delete the comments?

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -43,33 +65,46 @@ class BRBViewController: UIViewController {
...
+            self.accountViewController?.delegate = self
             addChildViewController(accountViewController)
             view.insertSubview(accountViewController.view, at: 0)
             accountViewController.view.snp.makeConstraints { make in
                 make.edges.equalToSuperview()
             }
             accountViewController.didMove(toParentViewController: self)
+
+            if let credentials = accountManager.getCredentials() {
+                navigationItem.title = "Hello, \(credentials.netid)"
```
danielVebman: Feature request: can we make this be their first name?

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -91,6 +126,13 @@ class BRBViewController: UIViewController {
         
         state = newState
     }
+    
+    func showErrorAlert(error: String) {
+        let errorAlert = UIAlertController(title: error, message: "Unable to fetch new BRB account data. Please check your connection and try again.", preferredStyle: UIAlertControllerStyle.alert)
+        errorAlert.addAction(UIAlertAction(title: "Okay", style: UIAlertActionStyle.destructive, handler: nil))
```
danielVebman: Why is this destructive? This is the (only) intended action. It should be `.default`.

Also, don't write out the whole class path if Swift can infer it. i.e.:

    errorAlert.addAction(UIAlertAction(title: "Okay", style: .default, handler: nil))

Eatery/AppDelegate.swift
```diff
@@ -12,9 +12,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var eateryTabBarController: EateryTabBarController!
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions:  [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-
         FirebaseApp.configure()
-
+
```
kevinchan159, line 17: delete

Eatery/Controllers/BRB/AboutTableViewController.swift
kevinchan159: if we don't end up cutting it, lets rename it to `atvc`

Eatery/Controllers/BRB/BRBAccountManager.swift
kevinchan159: i agree. lets move them into the classes that they're actually used in.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+// MARK: - Account Manager
+
+protocol BRBAccountManagerDelegate {
+    func brbAccountManagerDidFailToQueryAccount(with error: String)
+    func brbAccountManagerDidQuery(account: BRBAccount)
+}
+
+class BRBAccountManager {
+    
+    private var connectionHandler: BRBConnectionHandler!
```
kevinchan159: does this have to be `!` if we set it in the init?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+    func failedToLogin() -> Bool {
+        return loginCount > 1
+    }
+    
+    @objc func login() {
+        let javascript = "document.getElementsByName('netid')[0].value = '\(netid)';document.getElementsByName('password')[0].value = '\(password)';document.forms[0].submit();"
+        
+        evaluateJavaScript(javascript) { (result: Any?, error: Error?) -> Void in
+            if let error = error {
+                print(error)
```
whoiswillma: Remove `print` unless good reason to keep it?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+            block()
+        })
+    }
+}
+
+// MARK: - Account Manager
+
+protocol BRBAccountManagerDelegate {
+    func brbAccountManagerDidFailToQueryAccount(with error: String)
+    func brbAccountManagerDidQuery(account: BRBAccount)
```
whoiswillma: Similar comment as @danielVebman these names can probably be shorter and more swifty

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+            
+            if let account = account {
+                
+                let encoder = JSONEncoder()
+                if let encoded = try? encoder.encode(account) {
+                    let defaults = UserDefaults.standard
+                    defaults.set(encoded, forKey: "BRBAccount")
+                }
+                self.delegate?.brbAccountManagerDidQuery(account: account)
+
```
kevinchan159: delete newline on 260 and 267

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -8,16 +8,22 @@
...
         static let history = "history"
     }
     
-    private let account: BRBAccount
+    var account: BRBAccount
+    var delegate: BRBAccountViewControllerDelegate?
     
-    private var tableView: UITableView!
+    var tableView: UITableView!
+    var refreshControl: UIRefreshControl!
```
kevinchan159: can these be `private`?

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -43,6 +49,16 @@ class BRBAccountViewController: UIViewController {
         tableView.snp.makeConstraints { make in
             make.edges.equalToSuperview()
         }
+
+        refreshControl = UIRefreshControl(frame: .zero)
```
kevinchan159: you can just do `UIRefreshControl()`

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -43,6 +49,16 @@ class BRBAccountViewController: UIViewController {
...
+        refreshControl.tintColor = .white
+        refreshControl.addTarget(self, action: #selector(refreshBRBAccount), for: .valueChanged)
+        tableView.refreshControl = refreshControl
+    }
+
+    @objc private func refreshBRBAccount(_ sender: Any) {
+        delegate?.brbAccountViewControllerDidRefresh()
+        tableView.refreshControl!.endRefreshing()
```
kevinchan159: lets use `?` instead of `!`

Eatery/Controllers/BRB/AboutTableViewController.swift
whoiswillma: It's called `stvc` because the old name of the `AboutTableViewController` was the `SettingsTableViewController`.

Eatery/Controllers/BRB/BRBLoginViewController.swift
```diff
@@ -121,12 +80,13 @@ class BRBLoginViewController: UIViewController {
...
         setUpPasswordViews()
-        setUpSaveLoginInfoViews()
         setUpLoginViews()
+    }
 
-        if let (netid, password) = BRBAccountSettings.loadFromKeychain(), !netid.isEmpty, !password.isEmpty {
-            netidTextField.text = netid
-            passwordTextField.text = password
+    override func viewWillAppear(_ animated: Bool) {
+        if !(netidTextField.text?.isEmpty ?? true) && !(passwordTextField.text?.isEmpty ?? true) {
```
kevinchan159:
```
if let netID = netidTextField.text, 
  let password = passwordTextField.text, 
  !netID.isEmpty && !password.isEmpty {
  ...
}
```
might be a bit easier to read

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -110,42 +151,54 @@ extension BRBViewController: BRBLoginViewControllerDelegate {
...
     }
     
 }
 
+extension BRBViewController: BRBAccountViewControllerDelegate {
+    func brbAccountViewControllerDidRefresh() {
+        accountManager.queryBRBDataWithSavedLogin()
+        activityIndicator.startAnimating()
+    }
+}
```
kevinchan159: add newline after 177 and 181

kevinchan159 (commented):
almost there!

Eatery/Controllers/BRB/BRBAccountViewController.swift
whoiswillma: Can we get rid of this `refreshControl` variable since we assign it to `tableView.refreshControl`?

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -91,6 +126,13 @@ class BRBViewController: UIViewController {
         
         state = newState
     }
+    
+    func showErrorAlert(error: String) {
+        let errorAlert = UIAlertController(title: error, message: "Unable to fetch new BRB account data. Please check your connection and try again.", preferredStyle: UIAlertControllerStyle.alert)
```
whoiswillma: Same thing about the type inference thing here. We can use `.alert` instead of `UIAlertControllerStyle.alert`

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -15,26 +16,47 @@ class BRBViewController: UIViewController {
         case account(BRBAccount)
     }
     
-    private var connectionHandler: BRBConnectionHandler!
+    var accountManager: BRBAccountManager!
     private var requestStart: Date?
+    private var account: BRBAccount?
+    private var loggedIn = false
```
whoiswillma, line 19: Is it possible to have this be a computed property using `accountManager.stage`?

whoiswillma (changes requested):

Eatery/Controllers/BRB/BRBLoginViewController.swift
whoiswillma: Wait, forgive me if I'm misreading this in my tired state, but aren't we setting the `netidTextField.text = nil` if it's not nil, and `passwordTextField.text = nil` if it's not nil? So is it possible to remove this `if` branch if necessary and just put `netidTextField.text = nil` and `passwordTextField.text = nil`?

Eatery/Controllers/BRB/BRBAccountManager.swift
danielVebman: Maybe:

    brbAccountManager(didQuery account: BRBAccount)

Eatery/Controllers/BRB/BRBLoginViewController.swift
kevinchan159: haha you're right, good catch

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+//  Copyright © 2019 CUAppDev. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import WebKit
+
+struct HistoryEntry {
+
+    var description: String = ""
```
chalo2000: I'm pretty sure you can just remove the equals empty string because it just be initialized to that by default.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+private protocol BRBConnectionHandlerDelegate {
+
+    func brbConnectionHandlerDelegateDidRetrieve(sessionID: String)
+    func brbConnectionHandlerDelegateDidFailLogin(with error: String)
+
+}
+
+private class BRBConnectionHandler: WKWebView, WKNavigationDelegate {
+    
+    var stage: Stage = .loginScreen
```
chalo2000: how about putting some access modifiers on the rest of the variables to stay consistent and the functions below as well to explicitly state what they should be

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+    /// block after fetching the HTML for the page. Does not guarantee
+    /// Javascript will finish running before the block is executed.
+    func getStageAndRunBlock(block: @escaping () -> ()) {
+        getHTML(block: { (html: String) -> () in
+            if self.failedToLogin() {
+                self.stage = .loginFailed
+            } else if html.contains("<h2>Login OK</h2>") {
+                self.stage = .transition
+            } else if html.contains("<h1>CUWebLogin</h1>") {
+                if self.loginCount < 1 {
```
chalo2000: self.stage = self.loginCount < 1 ? .loginScreen : .loginFailed

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,277 @@
...
+    /// Gets the stage enum for the currently displayed web page and runs a
+    /// block after fetching the HTML for the page. Does not guarantee
+    /// Javascript will finish running before the block is executed.
+    func getStageAndRunBlock(block: @escaping () -> ()) {
+        getHTML(block: { (html: String) -> () in
+            if self.failedToLogin() {
+                self.stage = .loginFailed
+            } else if html.contains("<h2>Login OK</h2>") {
+                self.stage = .transition
+            } else if html.contains("<h1>CUWebLogin</h1>") {
```
chalo2000, line 158: i'm curious if there is a safer way to do this that doesn't involve checking the text field exactly (what if they change it to "CU Web Login" in the future?)

chalo2000 (dismissed):
very n o i c e

Eatery/Controllers/BRB/BRBViewController.swift
ReadeP: Not currently and adding that feature would take a lot of work in terms of the backend. But we can remove the numbers at the end of the user's name so that it looks cleaner and more personal by just showing their initials

danielVebman: That's an interesting idea. Maybe raise it with design?

ReadeP: Yeah I did, we are just going to set it to "Account Info" or something like that

ReadeP: The only issue I see with it is that since the user does not have to log in each time anymore, how do we go about telling them that they are looking at their account?

danielVebman: Why would we have to tell them they're looking at their account? Isn't that obvious? It's not like people are out here sharing phones.

ReadeP: I think it is an important confirmation to have

danielVebman: Doesn’t saying “Hello, bpd282” count as confirmation? Also, I would actually include the numbers. It might be confusing to a user to not see them given that they are so accustomed to it.

ReadeP: Yes, it would, but design said it looked weird and said to make it "Account Info". In that case I am thinking we should add a section title that shows the NetID or something like that

Eatery/Controllers/BRB/BRBViewController.swift
ReadeP: It is, but it introduces some issues when the user is already logged in and we query their cached account. Basically, with how the current enum Stages works, they would not be able to log out until the query is completed (if it ever is).

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -12,7 +12,8 @@ import SafariServices
 
 protocol AboutTableViewControllerDelegate: AnyObject {
 
-    func aboutTableViewControllerDidLogoutUser(_ stvc: AboutTableViewController)
+    func aboutTableViewControllerDidLogoutUser()
+    func aboutTableViewControllerDidTapBackButton()
```
danielVebman, line 16: I would remove all this spacing.

Eatery/Controllers/BRB/AboutTableViewController.swift
```diff
@@ -55,6 +56,14 @@ class AboutTableViewController: UITableViewController {
         tableView.register(LogoutTableViewCell.self, forCellReuseIdentifier: CellIdentifier.logout.rawValue)
     }
 
+    // Detects when user taps the 'Back' button in navigation bar
+    // Source: https://stackoverflow.com/a/32245222
+    override func willMove(toParentViewController parent: UIViewController?) {
+        if parent == nil {
```
danielVebman, line 62: Why do we check if the parent is nil?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,263 @@
...
+enum Stage {
+    case loginScreen
+    case transition
+    case loginFailed
+    case finished(sessionId: String)
+}
+
+// MARK: - Account Settings
+
+typealias LoginInfo = (netid: String, password: String)
```
danielVebman, line 27: I would make this a struct. What do the rest of y'all think?

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,263 @@
...
+        }
+        return (netid: netid, password: password)
+    }
+    
+}
+
+// MARK: - Connection Handler
+
+private protocol BRBConnectionHandlerDelegate {
+
```
danielVebman, line 58: Do we add these spaces for protocol declarations? I think they look bad.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,263 @@
...
+    case loginFailed
+    case finished(sessionId: String)
+}
+
+// MARK: - Account Settings
+
+typealias LoginInfo = (netid: String, password: String)
+
+private enum BRBAccountSettings {
+
```
danielVebman: Swift seems to favor the singleton pattern over a bunch of static methods. That would mean making the methods not static (i.e. instance), making the initializer private, and adding a `static let BRBAccountSettings shared = BRBAccountSettings()` to the top.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,263 @@
...
+    }
+    
+    func failedToLogin() -> Bool {
+        return loginCount > 1
+    }
+    
+    @objc func login() {
+        let javascript = "document.getElementsByName('netid')[0].value = '\(netid)';document.getElementsByName('password')[0].value = '\(password)';document.forms[0].submit();"
+        
+        evaluateJavaScript(javascript) { (result: Any?, error: Error?) -> Void in
```
danielVebman: I don't think you need to explicitly write the variable types for closure parameters. Also, we don't say `Void`, we say `()`.

i.e. say `(result, error) -> () in`.

Eatery/Controllers/BRB/BRBAccountManager.swift
```diff
@@ -0,0 +1,263 @@
...
+            } else {
+                if self.failedToLogin() {
+                    self.delegate?.brbConnectionHandler(didFailWith: "Incorrect netid and/or password")
+                }
+            }
+            self.loginCount += 1
+        }
+    }
+    
+    func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
```
danielVebman: Is that `!` still needed in the current Swift version?

danielVebman (commented):
Some small changes.

Eatery/Controllers/BRB/BRBAccountManager.swift
ReadeP: I did not have the spacing before, but Kevin told me to add it

danielVebman: Oh oops nvm then :)

Eatery/Controllers/BRB/BRBAccountManager.swift
whoiswillma: What makes you say we here? I'm unsure of what Transit does but I do not think this is standardized on Eatery

whoiswillma: What makes you say we here? I'm unsure of what Transit does but I do not think this is standardized on Eatery

whoiswillma: What makes you say we here? I'm unsure of what Transit does but I do not think this is standardized on Eatery

whoiswillma: :( Not sure why the comment was posted three times. I'll remove the other two

Eatery/Controllers/BRB/BRBAccountViewController.swift
```diff
@@ -8,14 +8,19 @@
...
+
 class BRBAccountViewController: UIViewController {
     
     private enum CellIdentifiers {
         static let balance = "balance"
         static let history = "history"
     }
     
-    private let account: BRBAccount
+    var account: BRBAccount
```
whoiswillma: Make this into `let`?

Eatery/Controllers/BRB/BRBViewController.swift
whoiswillma: What about making it be based off of `self.state`? Is that possible? 

If it's not, I think it would be best to clarify the relationship between `isLoading`, `state`, and `activityIndicator`

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -7,34 +7,53 @@
...
+        navigationItem.setLeftBarButton(activityItem, animated: true)
+
+        accountManager.delegate = self
+        accountManager.queryBRBDataWithSavedLogin()
+
+        if let account = accountManager.getCachedAccount() {
+            setState(.account(account))
+            activityIndicator.startAnimating()
+            loggedIn = true
+        }
```
whoiswillma, line 42: Move lines 30 to 42 into `viewDidLoad()`?

whoiswillma (approved):

Eatery/Controllers/BRB/BRBAccountManager.swift
danielVebman: Yeah what I said actually isn’t true. Maybe it’s a bad idea to review PRs at midnight 😬

Eatery/Controllers/BRB/BRBViewController.swift
ReadeP: We want to start querying when BRBViewController is initialized, not loaded. This means that even if the user stays on the eateries page, their account information will still be fetched in the background if that makes sense

whoiswillma: That makes sense. 

However, to make it more clean, I would say that `activityIndicator` should be a constant to avoid force unwraps.

Eatery/AppDelegate.swift
```diff
@@ -14,7 +14,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions:  [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
 
         FirebaseApp.configure()
-
+
```
kevinchan159, line 17: bump on deleting this

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -48,22 +67,25 @@ class BRBViewController: UIViewController {
...
-        
-        if case .finished = connectionHandler.stage {
+
+        if case .finished = accountManager.stage {
+            aboutVC.logoutEnabled = true
+        } else if loggedIn {
             aboutVC.logoutEnabled = true
         } else {
             aboutVC.logoutEnabled = false
         }
```
kevinchan159, line 77: can't we simplify this into one line:
```
aboutVC.logoutEnabled = accountManager.stage == .finished || loggedIn
```

Eatery/Controllers/BRB/BRBViewController.swift
```diff
@@ -99,53 +127,62 @@ extension BRBViewController: BRBLoginViewControllerDelegate {
...
-    
-    func aboutTableViewControllerDidLogoutUser(_ stvc: AboutTableViewController) {
+
+    func aboutTableViewControllerDidLogoutUser() {
         setState(.login)
-        
-        connectionHandler = BRBConnectionHandler()
-        connectionHandler.delegate = self
-        
+        navigationItem.title = "Account Info"
```
kevinchan159, line 176: we reference `"Account Info"`in `BRBViewController` multiple times, can we make it a class constant?

kevinchan159 (approved):
🙌 🙌 🙌

Eatery/Controllers/BRB/BRBAccountManager.swift
ReadeP: What's wrong with NSString? 🤡 🫵

chalo2000: yuck indeed

kevinchan159: 😆

whoiswillma: IMG_0044

PR #250, opened 2019-11-21
Add Image Caching for Eateries

whoiswillma:
## Overview

Use Kingfisher's image caching to save bandwidth. Also add `UICollectionView` prefetching.

## Test Coverage

I used the network link conditioner to artificially limit network performance to simulate what it will be like when we switch to higher quality images.


kevinchan159 (approved):
CLEANNN

ReadeP (approved):
👌👌👌

PR #251, opened 2019-11-23
Refactor OnboardingViewController

ReadeP:
## Overview
In this PR I worked on refactoring `OnboardingViewController` by creating two subclasses `OnboardingInfoViewController` and `OnboardingLoginViewController`. Each manages its own contentView which can can either contains an imageView or a loginStackView. This prevents the need to access the stackView in `OnboardingViewController`

## Changes Made
- Made stackView private in `OnboardingViewController`
- Created `OnboardingInfoViewController`
- Created `OnboardingLoginViewController`

## Test Coverage
I tested this feature manually.

## Next Steps (delete if not applicable)
Work on adding placement animations using Lottie.

## Related PRs or Issues (delete if not applicable)
Will create a future PR that includes animations


Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+        super.viewDidLoad()
+
+        setUpStackView()
+        setUpImageView()
+        setUpButton()
+
+        stackView.layoutIfNeeded()
+        contentView.snp.makeConstraints { make in
+            make.width.equalToSuperview()
+            make.height.equalTo(stackView.frame.height)
```
whoiswillma: Why use this instead of `make.height.equalTo(stackView)`?

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
```diff
@@ -55,18 +56,33 @@ class OnboardingLoginViewController: OnboardingViewController {
...
         setUpNetidViews()
         setUpPasswordViews()
+        setUpButton()
         setUpSkipButton()
+
+        stackView.layoutIfNeeded()
+        contentView.snp.makeConstraints { make in
+            make.width.equalToSuperview()
+            // 52.5 is the height of the errorView
+            make.height.equalTo(stackView.frame.height + 52.5)
```
whoiswillma: Once more about the `make.height.equalTo(stackView.height)`

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+import UIKit
+
+class OnboardingInfoViewController: OnboardingViewController {
+
+    private let stackView = UIStackView()
+    private let imageView = UIImageView()
+    private let nextButton = UIButton()
+    private var onboardingImage = UIImage()
+
+    init(title: String, subtitle: String, image: UIImage!) {
```
whoiswillma: Change `image: UIImage?`. I think making it a normal optional is fine. If the image that's passed in is nil then `imageView` can display nothing.

whoiswillma (changes requested):

whoiswillma:
I forgot to mention: does this fix the issue where the login button is not tappable?

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
danielVebman: Why do you have `UIImage!`? Is the image optional? In that case say `UIImage?`. Also, you can remove the assignment on line 16, make it a `let` constant, and move `self.onboardingImage = image` to line 18 right before the super call. 

That is, line 16 would be `private let onboardingImage: UIImage`, and you move line 20 before line 19.

Eatery/Onboarding/Controllers/OnboardingViewController.swift
```diff
@@ -14,24 +14,21 @@ protocol OnboardingViewControllerDelegate {
...
     private let titleLabel = UILabel()
-    private var imageView: UIImageView!
-    private let nextButton = UIButton()
+    private let subtitleLabel = UILabel()
 
     private var onboardingTitle: String!
     private var onboardingSubtitle: String!
-    private var onboardingImage: UIImage?
+
+    let contentView = UIView()
```
danielVebman, line 26: Can this be private?

Eatery/Onboarding/Controllers/OnboardingViewController.swift
```diff
@@ -14,24 +14,21 @@ protocol OnboardingViewControllerDelegate {
...
-    let subtitleLabel = UILabel()
-    
+    private let stackView = UIStackView()
     private let titleLabel = UILabel()
-    private var imageView: UIImageView!
-    private let nextButton = UIButton()
+    private let subtitleLabel = UILabel()
 
     private var onboardingTitle: String!
     private var onboardingSubtitle: String!
```
danielVebman: Like my first comment, remove the `!` and move lines 30 and 31 before the super call.

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
ReadeP: If we are adding  52.5 we have to specify .frame.height

Eatery/Onboarding/Controllers/OnboardingViewController.swift
ReadeP: Can't be since we are accessing it in OnboardingViewController's subclasses

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
ReadeP: Actually we do not even need to add the 52.5

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        setUpStackView()
+        setUpImageView()
+        setUpButton()
+
+        stackView.layoutIfNeeded()
```
whoiswillma: Is this necessary?

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
```diff
@@ -55,18 +56,32 @@ class OnboardingLoginViewController: OnboardingViewController {
         setUpErrorView()
         setUpNetidViews()
         setUpPasswordViews()
+        setUpButton()
         setUpSkipButton()
+
+        stackView.layoutIfNeeded()
```
whoiswillma: Same thing here: needed?

whoiswillma (changes requested):
Looks really good! However, I still don't think this solves the issue of the software keyboard blocking the next button. Do you have any plans as how to solve that?

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
ReadeP: Not anymore

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+            make.height.equalTo(stackView)
+        }
+    }
+
+    private func setUpStackView() {
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.alignment = .center
+        stackView.spacing = 40
+        contentView.addSubview(stackView)
```
kevinchan159, line 45: newline after here

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+        stackView.snp.makeConstraints { make in
+            make.center.equalToSuperview()
+            make.width.equalToSuperview()
```
kevinchan159: you can do `make.center.width.equalToSuperview()`

kevinchan159 (approved):
nice work!

Eatery/Onboarding/Controllers/OnboardingInfoViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    private func setUpStackView() {
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.alignment = .center
+        stackView.spacing = 40
+        contentView.addSubview(stackView)
+
+        stackView.snp.makeConstraints { make in
+            make.center.equalToSuperview()
```
whoiswillma: Can get rid of this constraint

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
```diff
@@ -55,18 +56,35 @@ class OnboardingLoginViewController: OnboardingViewController {
...
         setUpPasswordViews()
+        setUpButton()
         setUpSkipButton()
+
+        contentView.snp.makeConstraints { make in
+            make.width.equalToSuperview()
+            make.height.equalTo(stackView)
+        }
+
+        NotificationCenter.default.addObserver(self, selector: #selector(self.keyboardWillShow), name: NSNotification.Name.UIKeyboardWillShow, object: nil)
```
whoiswillma: Can use type inference to shorten this line to `NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillShow), name: .UIKeyboardWillShow, object: nil)`

It might not be possible to shorten to `.UIKeyboardWillShow` since Eatery uses an older version of Swift, not sure

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
```diff
@@ -55,18 +56,35 @@ class OnboardingLoginViewController: OnboardingViewController {
...
 
     private func setUpStackView() {
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.alignment = .center
+        stackView.spacing = 40
+        contentView.addSubview(stackView)
+
+        stackView.snp.makeConstraints { make in
+            make.center.equalToSuperview()
```
whoiswillma: `make.center.width.equalToSuperview()` is a bit cleaner

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
```diff
@@ -223,6 +246,30 @@ class OnboardingLoginViewController: OnboardingViewController {
         }
     }
 
+    @objc func keyboardWillShow(notification: NSNotification) {
+        guard let userInfo = notification.userInfo else { return }
+
+        guard let keyboardSize = userInfo[UIKeyboardFrameEndUserInfoKey] as? NSValue else { return }
```
whoiswillma: Can combine both of these guard statements into one by doing
```
guard let userInfo = ...,
    let keyboardSize = ... else { return }
```

Eatery/Onboarding/Controllers/OnboardingLoginViewController.swift
```diff
@@ -223,6 +246,30 @@ class OnboardingLoginViewController: OnboardingViewController {
...
+
+        let keyboardFrame = keyboardSize.cgRectValue
+
+        if self.view.frame.origin.y == 0 {
+            self.view.frame.origin.y -= keyboardFrame.height * 2 / 3
```
whoiswillma: This is a bit hacky; Its strange for a view controller to adjust its view's position inside its superview.

I think a way to avoid doing this would be to add keyboard logic into the `OnboardingViewController` and have it adjust the entire stackview upwards when it detects the keyboard.

whoiswillma (changes requested):

Eatery/Onboarding/Controllers/OnboardingPrivacyStatementViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        setUpCloseButton()
+    }
+
+    private func setUpCloseButton() {
+        let closeButton = UIButton()
+        closeButton.setImage(UIImage(named: "closeIcon.png"), for: .normal)
```
whoiswillma: Since `closeIcon` is in the assets catalog I think the `.png` should be removed.

Eatery/Onboarding/Controllers/OnboardingPrivacyStatementViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+        closeButton.addTarget(self, action: #selector(dismissSelf), for: .touchUpInside)
+        view.addSubview(closeButton)
+
+        closeButton.snp.makeConstraints { make in
+            make.width.height.equalTo(50)
+            make.topMargin.equalToSuperview().offset(10)
```
whoiswillma: `make.topMargin.rightMargin.equalToSuperview().inset(10)`

whoiswillma (approved):

PR #253, opened 2019-12-02
Add Popular Times Response

whoiswillma:
## Overview

Users can report how long they actually waited for food at an eatery. Responses are recorded in Firebase. Users can only submit one response per hour per eatery and only while the eatery is open. 

The popular times response is initially a call-to-action that then expands into a full prompt. After the user responds, the prompt fully disappears.

## Changes Made

I added `PopularTimesResponse` class that rate limits the responses and actually logs the response in Firebase. The biggest downside is that `PopularTimesResponse` is a hodge-podge of features that touches logging, user defaults, and campus eatery. However, I think it's also the cleanest way to do it. 

The other changes are purely UI related. The `PopularTimesView`manages the `PopularTimesAccuracyPrompt` view, which is a display-only view. 

## Test Coverage

Entirely manual testing. Over the course of development I have tested that the one-hour limit eventually expires.

## Screenshots

<details>

  <summary>Popular Times Response default actions</summary>

RPReplay_Final1575302139 MP4

</details>

<details>

  <summary>Example of Popular Times response</summary>

Times are in the format agreed upon, rounded down to the nearest half hour. For context I took this screenshot at 10:58.

Screen Shot 2019-12-02 at 10 58 56

</details>


kevinchan159 (approved):
looks good to me!

ReadeP (approved):

PR #259, opened 2020-02-04
Fixes Apollo Build Phase Script & updates README with updated onboarding instructions

kevinchan159:
## Overview

I made changes to the `Apollo` build script phase to simplify setup of Apollo for new users and also updated the README with new onboarding instructions (some of the instructions were also outdated, i.e. we don't need `Keys.plist` anymore).

## Changes Made

Currently both `Eatery` and `Eatery Watch App Extension` relies on `Apollo` which means that both of these two targets require an `schema.json` to generate their necessary `API.swift` files. The old Apollo Build Script phase checked for `schema.json` in the `$SRCROOT/$TARGET_NAME` directory which means that users would need the `schema.json` file in both the `Eatery` and `Eatery Watch App Extension` directories which is unnecessary because the `schema.json` files are the same anyways.

To fix this, I changed it so that when the Apollo build phase script is run, it just looks for `schema.json` inside `$SRCROOT/schema.json` which means users will just need to have the `schema.json` file downloaded inside the root project directory.

## Test Coverage

Ran the app and made sure it builds

### DIFF
--- a/Eatery.xcodeproj/project.pbxproj
+++ b/Eatery.xcodeproj/project.pbxproj
@@ -1094,7 +1094,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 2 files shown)


usubzero (approved):
Verified changes result in successful build.

whoiswillma (approved):
LGTM!

ywsang (approved):
lgtm

PR #276, opened 2020-03-04
Fix state transition visual bugs

whoiswillma:
## Overview

Fix bugs related to state transitions in the Eateries View Controller. 
1. Manually refreshing Eateries would cause the Eateries to flash, then animate in the remaining Eateries
2. Occasionally, during animation, the Eatery cells would snap to their final position.

## Changes Made

1. Caused by a reload after fading in the collection view that cancelled the fade in animation. 

   I moved the fade in animation to be after the reload. 
2. Caused by the update timer triggering while the eateries were animating.

   I invalidate the update timer when state transitions occur, and only activate it when we are presenting Eateries

## Test Coverage

Manual testing on device.


kevinchan159 (approved):
nice

PR #278, opened 2020-04-15
Implement app icon switcher

whoiswillma:
## Overview

Add ability to switch app icon from the settings page. 

Let me know if I should change the text for the winter icon.

## Changes Made

* Add winter icon to `Resources/AppIconWinter`
* Add section to About View Controller for switching icons 

## Test Coverage

Tested on a physical iPhone on iOS 13

## Screenshots

<details>

  <summary>Settings Page</summary>

IMG_9FE672BBD815-1

  
  

</details>


kevinchan159 (approved):
🔥 
i would just message Amanda for a final approval before shipping

danielVebman (approved):
Looks good. I wonder from a UX perspective though if it makes sense to give users an option. An alternative would be to automatically switch the icon based on the month or to have it determined by Firebase remote config / equivalent.

Issue #273, opened 2020-02-29
Improve Search


whoiswillma:
* See previous searches
* Put the search bar into the navigation bar, a la Settings, Health, every other app that uses a search bar.
* Instead of hackily updating the eateries collection view, we should use a `UISearchController`
* Distinguish between searching Eateries and menu items: currently the two are mixed in an awkward way

PR #281, opened 2020-05-03
[Improved Search][Part 3] Integrate Search Results

whoiswillma:
## Overview

Integrate search results with the rest of Eatery, i.e. tap on a search result to open it.

## Changes Made

Connect the `CampusEateriesSearchViewController` to `CampusEateriesViewController` using a delegate. 

`UISearchController` behaves weirdly with how we're doing navigation bars. The search controller insists that there's a navigation bar (so it can show the search bar), but when we push a menu view controller we want the navigation bar to disappear so we can present our own "fake" transparent navigation bar. The `UISearchController` wins out and the navigation bar is presented in addition to our navigation bar. 

This all means I have to dismiss the `UISearchController` before I can present the search result. This is done by storing the `selectedSearchResult`, then waiting for a call from `UISearchController` when its done dismissing. 

Some other random things I did:
* Search filtering was readded (oops)
* Slight reworking of `FilterBar` so `selectedFilters` is the source of truth. 

## Test Coverage

Tested on a physical iPhone 11 running iOS 13

## Next Steps

Collegetown 

## Related PRs or Issues

#273 

## Screenshots (delete if not applicable)

<details>

  <summary>Screen Shot Name</summary>

RPReplay_Final1588518981

  

</details>

### DIFF
--- a/Eatery/Controllers/Eateries/Campus/CampusEateriesSearchViewController.swift
+++ b/Eatery/Controllers/Eateries/Campus/CampusEateriesSearchViewController.swift
@@ -47,7 +47,9 @@ private class SearchResultsManager {
 
     private func eaterySearchResults(searchText: String) -> [SearchResult] {
         eateries.filter { eatery in
-            matches(eatery.name, searchText) || matches(eatery.displayName, searchText)
+            matches(eatery.name, searchText)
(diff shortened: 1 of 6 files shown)


usubzero (approved):
Looks great!

PR #283, opened 2020-08-17
Miscellaneous bug fixes in preparation for Eatery 2.9

whoiswillma:
## Overview

Some changes / bug fixes for v2.9 release

 * Hide watch redesign
 * Remove meal swipes from account info
 * Pop view controller when eatery tab bar item tapped
 * Persist hours-this-week and popular-times expanded/collapsed state
 * Fix AppDev logo appearing in search bar on first launch / refactored the presentation of the AppDev logo

More recent commits

 * Add build script to automatically set the build number 
 * Remove External Beta, Internal Beta and App Store configurations; Also remove Eatery Production - the difference between these was never clear to me at least, and the only distinction I could gather was that App Store was used with Eatery Production, which we used while archiving the project. So they added extra complexity for no discernible reason


Eatery/Views/Eateries/CampusMenuInfoView.swift
```diff
@@ -115,16 +121,19 @@ class CampusMenuInfoView: UIView, DynamicContentSizeView {
...
 
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
     override func didMoveToWindow() {
+        super.didMoveToWindow()
+
         // Defer layout until view has been added to view hierarchy
         // to avoid unsatisfiable layout constraints
```
ywsang: do these comments still apply?

Eatery/Controllers/EateryTabBarController.swift
```diff
@@ -54,7 +54,11 @@ extension EateryTabBarController: UITabBarControllerDelegate {
     func tabBarController(_ tabBarController: UITabBarController, shouldSelect viewController: UIViewController) -> Bool {
         if selectedViewController === viewController,
             let shared = viewController as? EateriesSharedViewController {
-            shared.activeViewController.scrollToTop(animated: true)
+            if shared.activeNavigationController.viewControllers.count > 1 {
```
ywsang, line 57: just wanna make sure I'm understanding what's happening here -- will admit I'm not very familiar with the eatery codebase/architecture. 

if there's currently an eatery detail view controller showing, and the user taps on the leftmost tab (the one for eateries), it'll pop back to the list of eateries, maintaining the scroll position that the user was in before selecting that eatery? otherwise, the list of eateries will scroll to the top of the list?

ywsang (approved):
had a question or two just to clarify my understanding, but otherwise lgtm! 
nice work on the logo positioning -- looks super nice

Eatery/Controllers/EateryTabBarController.swift
whoiswillma: That's the intended behavior. The music app does it this way for example

Eatery/Views/Eateries/CampusMenuInfoView.swift
whoiswillma: Oops, resolved

Eatery/Controllers/Eateries/EateriesViewController.swift
```diff
@@ -841,19 +843,41 @@ extension EateriesViewController: UIScrollViewDelegate {
...
-        let offset = collectionView.contentOffset.y + collectionView.adjustedContentInset.top
+        let navBarHeight = navigationController?.navigationBar.frame.height ?? 0.0
+        let y = appDevLogoYPosition(navBarHeight: navBarHeight)
+
+        let navBarWidth = navigationController?.navigationBar.frame.width ?? 0.0
+        let centerOfRightNavigationBarItem = navBarWidth - 40
+        let x = centerOfRightNavigationBarItem - EateriesViewController.appDevLogoSize / 2
+
+        appDevLogo.transform = CGAffineTransform(translationX: x, y: y)
+        appDevLogo.alpha = min(0.9, y / 60 - 2)
```
ywsang: could we add a comment here explaining what the min comparison/calculation is for?

ywsang (approved):
lgtm! just one small comment. 

for the different schemes -- you definitely have more knowledge than me about eatery has been using schemes in the past, so if you haven't had to use the two different beta configs in the past I think it's fine to get rid of them. everything looks fine to me.

PR #288, opened 2020-08-23
Fix style issue in `Eatery/Analytics` and `Eatery/Controllers/BRB`

whoiswillma:
## Overview

Fix SwiftLint violations in `Eatery/Analytics` and `Eatery/Controllers/BRB`


Eatery/Controllers/BRB/BRBConnectionHandler.swift
```diff
@@ -1,165 +0,0 @@
-import UIKit
```
ywsang, line 1: is this just not used anymore, or moved into the account manager?

ywsang (approved):
lgtm!

Eatery/Controllers/BRB/BRBConnectionHandler.swift
whoiswillma: It got moved into the account manager. It was deleted from the Xcode project, but not from the file system.

PR #290, opened 2020-08-24
BRB: Fix crash caused by uninitialized view

### LINKED ISSUE #287: Previous incorrect login causes app to crash
It seems like a previous incorrect netID or password triggers the `errorDescription` in the `BRBLoginViewController` to be set on subsequent launches of the app. This results in an "Unexpectedly found nil" error on line 32 on the `BRBLoginViewController` if the login tab is not selected within the first few seconds after opening the app -- probably due to the `errorView` not being initialized yet.

whoiswillma:
Fixes #287

### DIFF
--- a/Eatery/Controllers/BRB/BRBLoginViewController.swift
+++ b/Eatery/Controllers/BRB/BRBLoginViewController.swift
@@ -31,7 +31,10 @@ class BRBLoginViewController: UIViewController {
     private var errorView: BRBLoginErrorView!
     var errorDescription: String? {
         get { errorView.errorLabel.text }
-        set { errorView.errorLabel.text = newValue }
+        set {
+            loadViewIfNeeded()
+            errorView.errorLabel.text = newValue
+        }
     }
 
     private var netidPrompt: UILabel!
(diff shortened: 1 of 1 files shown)


ywsang (commented):
curious about the tradeoffs between doing all the work in the background vs. doing it when the account tab is selected. 
in case the user never actually goes to the account tab in the app, is there a significant amount of work being done to get the account info, load the views, etc. that causes a significant slowdown in the performance of the app?
if the word isn't done beforehand, is there a lag in the login info if the account tab is selected?

whoiswillma:
Sure, I can make it lazy-load, but frankly I didn't see any noticeable performance issues on an admittedly newer iPhone. We're not doing anything synchronous in `viewDidLoad` besides initializing a dozen or so views.

ywsang:
I’m also running on a relatively newer iPhone, so I can’t really speak to performance on older devices either — but if there haven’t been issues/complaints about performance then I think it’s fine to leave it as is.

ywsang (approved):

PR #291, opened 2020-08-25
eatery: Change EateryType.dining.rawValue to 'dining room'

whoiswillma:
Change `EateryType.dining.rawValue` to `"dining room"` to match with Cornell's new description.

I think we'll want a more robust way to do this in the future. Could this be solved on the backend, or with regexes instead of exact string matches?

### DIFF
--- a/Shared/Models/Eatery.swift
+++ b/Shared/Models/Eatery.swift
@@ -37,9 +37,9 @@ enum PaymentMethod: String, Codable {
 }
 
 /// Different types of eateries on campus
-enum EateryType: String, Codable{
+enum EateryType: String, Codable {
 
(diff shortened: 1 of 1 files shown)


Shared/Models/Eatery.swift
```diff
@@ -39,7 +39,7 @@ enum PaymentMethod: String, Codable {
 /// Different types of eateries on campus
 enum EateryType: String, Codable{
```
ywsang: sorry tiny nit: can we add a space before the `{`?

ywsang (approved):
Have the fields changed frequently in the past? I would lean away from using regexes, just because it seems like it could be reasonably buggy, especially if Cornell is changing it pretty frequently.

PR #296, opened 2020-09-02
Add TestFlight scheme with Wormholy

whoiswillma:


Podfile
```diff
@@ -23,6 +23,7 @@ target 'Eatery' do
     pod 'NVActivityIndicatorView'
     pod 'SnapKit'
     pod 'lottie-ios'
+    pod 'Wormholy', :configurations => ['Debug', 'TestFlight']
```
ywsang, line 26: what's the difference between the `Debug` and `TestFlight` configurations? based on the `.xcsheme` files in the commit, it looks like the testflight config isn't used?

ywsang (changes requested):
one small question

Podfile
whoiswillma: I dug around and found the TestFlight .xcscheme file inside xcuserdata, whereas all the rest are in xcshareddata? I'm not sure what's up with that, but I recreated the scheme and that seemed to have moved it

whoiswillma: I had to create a TestFlight scheme (instead of just using Debug) because FLEX uses UIWebView which Apple no longer accepts

ywsang (approved):
ah, got it! looks good

PR #300, opened 2020-09-28
Add Eatery Exceptions View

whoiswillma:
## Overview

Add exceptions view and connect it to the backend. We fade in the exceptions view after the view controller has appeared so that the blur can be computed by UIKit. Otherwise, the blur is based off a grey background that isn't actually there and it's jarring to see the blur change color.

Since in general we can get any text from the backend, the exception view is properly constrained to avoid falling off the right-hand side of the view.

Also do some cleanup work with `UIColor` and the Xcode project.

## Changes Made

  - Add exceptions to Eatery
  - Disable targeting arm64 so Eatery can run in the simulator 
  - Remove `colorFromCode` and use a conv. initializer instead to be more swifty
  - Add `exceptions` to graphql query and update `API.swift`

## Test Coverage

Manual testing on iPhone 11

## Next Steps (delete if not applicable)

Get button

## Screenshots (delete if not applicable)

<details>

  <summary>Screen Shot Name</summary>

IMG_2561
IMG_2560
IMG_2559

  
  

</details>


Eatery/Views/Eateries/EateryExceptionsView.swift
```diff
@@ -0,0 +1,64 @@
...
+            make.width.height.equalTo(16)
+        }
+
+        label.font = .systemFont(ofSize: 12)
+        label.textColor = .white
+        insertSubview(label, aboveSubview: blurView)
+        label.snp.makeConstraints { make in
+            make.leading.equalTo(imageView.snp.trailing).offset(4)
+            make.top.bottom.equalToSuperview().inset(4)
+            make.trailing.equalToSuperview().inset(8)
```
ywsang, line 44: don't know what the convention has been in eatery thus far, but with the repeat of the constant padding values I could see a case for making constants and naming them so they describe what they are -- but really just a nitpicky thing. I'd just stick to whatever matches the current codebase does

ywsang (approved):
🚀  looks super clean! `UIColor` file also looks a lot cleaner

Eatery/Views/Eateries/EateryExceptionsView.swift
whoiswillma: Current codebase uses literals

PR #306, opened 2020-11-06
EateriesViewController: Fix pull-to-refresh too sensitive

whoiswillma:
## Overview

Not sure if this is iOS 13 only, but the refresh control is super sensitive when you first launch the app. This PR fixes that

## Changes Made

Add the refresh control to the collection view after the view has been added to the window.

## Test Coverage

Physical iPhone 11


ywsang (approved):
hmm interesting

spd75 (approved):
All looks good

PR #307, opened 2021-01-26
ExpandedMenus

spd75:
## Overview
I implemented ExpandedMenus into Eatery.  This update provides a much more interactive interface to non-DiningHall eateries (such as Bear Necessities, Libe Cafe, Trillium, etc.).  The feature provides users with a full menu of all the items offered at specific eateries.  It then sorts these menu items into categories, such as "pizza", "burgers", "sub-sandwiches", etc.  This ExpandedMenus feature also allows for sorting of these menu items by price.  I attached the Figma file below to give a better visual of how this feature should look and work. 

https://www.figma.com/file/Kb2QmOQhFPJZzncFJSXZZR/Caf%C3%A9-Menus?node-id=38%3A1414

*Please note that this feature required A LOT of mathematical calculations to calculate scroll position, where to scroll, how to scroll the header tab bar with the rest of the screen, etc.  This makes my code look messy, especially since I tried to follow XCode's 120 character-per-line rule.  You can find a lot of this messiness in `ScrollablePageViewController`. Feel free to give suggestions on how to make my code look cleaner.

## Changes Made
The main change I made was within `CampusMenuViewController.swift`.  Here, depending on the `eateryType` property of the eatery, either a `TabbedPageViewController` or a `ScrollablePageViewController` is rendered.  

I created `ScrollablePageViewController.swift`, which contains all the code for expandedMenus.  From here, I created the following subViewControllers and subViews:
- `ScrollableTextTabBarControl`: the `UIView` that handles the scrollable tab bar in the ExpandedMenu header
- `CampusEateryExpandedMenuViewController`: a sub `UIViewController` which contains a certain category of the ExpandedMenu - many of these put together create the entire ExpandedMenu.  This also contains some sub `UIViews` which should be self-explanatory.
- `ExtendedFilterbutton`: the customized `UIButton` that serves as the filter button

I also had to change some things related to networking to get the correct data from the backend.

## Test Coverage
Since this was primarily a visual feature added, I didn't have to develop many test cases.  However, I did test multiple variations of user input (scroll-filter-scroll, filter-filter-filter, scroll-changetabs-changetabs, etc.) and everything checked out. 

My biggest issue was adjusting for screen size.  Because the newer iPhones, such as the iPhone 11 and iPhone 12 have large edge insets, the scrolling of the main tab bar for ExpandedMenus wouldn't work properly on these screens.  To account for this, I had to change a few things in the `ImageParallaxScrollViewController` and create a `scrollOffset` variable that would always account for these edge insets.  I then tested all screen sizes multiple times. 

## Next Steps
The next step is very simple, consisting of only adding a Cornell background image at the bottom of the ExpandedMenu.  Zain provides a visual example of this in the last diagram of the Figma file, which is under the "Overview" folder.  For this PR, I only implemented a blank gray screen.

## Related PRs or Issues
The main issue with this feature is that a lot of the data is hardcoded from the backend.  The current data provides the menu for every non-DiningHall eatery and assigns it to each eatery (including DiningHall eateries), rather than individualizing each menu.  This makes load times take much longer because a lot of unnecessary data has to be fetched and rendered into their respective UIViews.  Ideally, this semester, the backend team could fix that.

## Screen-Recordings

<details>

  <summary>Video Demonstration on iPhone 8 -- With GET Button</summary>

https://user-images.githubusercontent.com/56354603/105899819-29e7ae00-5fe9-11eb-829c-5d099be6c55e.mov

https://user-images.githubusercontent.com/56354603/105900238-c27e2e00-5fe9-11eb-8c08-40566933c8af.mov

  
  

</details>


Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+//  Copyright © 2020 Cornell AppDev. All rights reserved.
+//
+
+import UIKit
+
+class ScrollableViewController: UIViewController {
+
+    private var eatery: CampusEatery!
+    private var viewControllers: [CampusEateryExpandedMenuViewController]!
+    private var allItems: [ExpandedMenu.Item]!
```
whoiswillma: Can we make these into `private let`'s

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+    private var menuItemsView: UIView!
+    private var filterButton: ExtendedFilterButton!
+    private var filterLabel: UILabel!
+    private var headerContainer: UIView!
+    private var separatorView: SeparatorView!
+
+    let tabBarPadding: CGFloat = 16
+    var scrollView: UIScrollView?
+    var scrollOffset: CGFloat = 0
+    var fillerSpaceHeight: CGFloat = 0
```
whoiswillma: Is it possible to make these private?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+            let checkHeight = reducedScreenSize - (menuItemsStack.frame.maxY - finalView.frame.minY)
+            fillerSpaceHeight = checkHeight > 0 ? checkHeight : 0
+        }
+
+        self.scrollView?.contentSize.height += fillerSpaceHeight
+    }
+
+    private func createTabBarStack() {
+        headerContainer = UIView()
+        headerContainer.layer.zPosition = 3
```
whoiswillma: What does this line do? Not too sure how layer `zPosition` works

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+            let finalView = originalViews[originalViews.count - 1]
+            let checkHeight = reducedScreenSize - (menuItemsStack.frame.maxY - finalView.frame.minY)
+            fillerSpaceHeight = checkHeight > 0 ? checkHeight : 0
+        }
+
+        self.scrollView?.contentSize.height += fillerSpaceHeight
+    }
+
+    private func createTabBarStack() {
+        headerContainer = UIView()
```
whoiswillma: How do you feel about pulling out this `headerContainer` into its own class?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+        tabBar.snp.makeConstraints { make in
+            make.height.equalTo(40)
+        }
+
+        tabBarStack = UIStackView(arrangedSubviews: [headerContainer, tabBar])
+        tabBarStack.axis = .vertical
+    }
+
+    private func createMenuStack() {
+        menuItemsView = UIView()
```
whoiswillma: May be worthwhile to pull this out into its own class as well

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+    private func createHighToLowStack() {
+        let reversedItems = Array(allItems.reversed())
+        let sortedVC = CampusEateryExpandedMenuViewController(eatery: eatery, category: "", menu: reversedItems)
+
+        menuItemsHighToLowStack = UIStackView(arrangedSubviews: [sortedVC.view])
+        menuItemsHighToLowStack.axis = .vertical
+        menuItemsView.addSubview(menuItemsHighToLowStack)
+        menuItemsHighToLowStack.isHidden = true
+
+        self.menuItemsHighToLowStack.snp.makeConstraints { make in
```
whoiswillma:
```suggestion
        menuItemsHighToLowStack.snp.makeConstraints { make in
```

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+
+        menuItemsStack.snp.makeConstraints { make in
+            make.top.bottom.equalToSuperview()
+            make.leading.trailing.equalToSuperview().inset(5)
+        }
+
+    }
+
+    private func createHighToLowStack() {
+        let reversedItems = Array(allItems.reversed())
```
whoiswillma: I think this array is being reversed because it's high to low? Either way could you add a comment explaining why the items are reversed?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
whoiswillma: It might also be good to add a few doc comments to these properties. Is `allItems` supposed to be in sorted order?

Eatery/Controllers/Eateries/CampusMenuViewController.swift
```diff
@@ -156,6 +174,31 @@ class CampusMenuViewController: EateriesMenuViewController {
         }
     }
 
+    private func addExtendedMenuViewController() {
+        view.backgroundColor = .wash    // Changing background color here simplifies ScrollableViewController
+
+        if eatery.orderedExpandedCategories.count == 0 {
```
whoiswillma:
```suggestion
        if eatery.orderedExpandedCategories.isEmpty {
```

Eatery/Controllers/Eateries/CampusMenuViewController.swift
```diff
@@ -156,6 +174,31 @@ class CampusMenuViewController: EateriesMenuViewController {
...
+            let extendedMenuVC = CampusEateryExpandedMenuViewController(eatery: eatery, category: category, menu: menu)
+            viewControllers.append(extendedMenuVC)
+            menuItems.append(contentsOf: menu)
+        }
+
+        scrollableVC = ScrollableViewController(eatery: eatery, viewControllers: viewControllers, items: menuItems)
+        addChildViewController(scrollableVC!)
+        addToStackView(scrollableVC!.view)
+        scrollableVC!.didMove(toParentViewController: self)
+        scrollableVC!.view.hero.modifiers = createHeroModifiers(.fade, .translate)
```
whoiswillma: Can we avoid some of these force unwraps by using `let scrollableVC = ...` then assigning to self after?
```suggestion
        let scrollableVC = ScrollableViewController(eatery: eatery, viewControllers: viewControllers, items: menuItems)
        addChildViewController(scrollableVC)
        addToStackView(scrollableVC.view)
        scrollableVC.didMove(toParentViewController: self)
        scrollableVC.view.hero.modifiers = createHeroModifiers(.fade, .translate)
        self.scrollableVC = scrollableVC
```

Eatery/Utils/Sort.swift
```diff
@@ -110,6 +110,56 @@ struct Sort {
         return eatery.sorted(by: sortByHoursClosure)
     }
 
+    // Merge Sort code inspired by https://github.com/raywenderlich/swift-algorithm-club/tree/master/Merge%20Sort
+
+    static func expandedMenuMergeSort(_ array: [ExpandedMenu.Item]) -> [ExpandedMenu.Item] {
```
whoiswillma: Haven't dug into it too deep, but what's the reasoning for creating your own sort function over using `array.sort` and giving a custom `isLessThan` closure?

Eatery/Views/Eateries/ExpandedMenuRow.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    var itemLabel: UILabel!
+    var priceLabel: UILabel!
+    var item: ExpandedMenu.Item!
+
+    let padding: CGFloat = 15
+
+    let leadPadding: CGFloat = 15
+    let trailPadding: CGFloat = -15
+    let lineOffset: CGFloat = 2
```
whoiswillma: Can some of these be made private, if possible?

Eatery/Views/Eateries/ExpandedMenuRow.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    init(item: ExpandedMenu.Item, hasSeparator: Bool) {
```
whoiswillma:
```suggestion
    init(item: ExpandedMenu.Item) {
```

Eatery/Views/Eateries/ExtendedFilterButton.swift
```diff
@@ -0,0 +1,166 @@
...
+    var topRect: UIView!
+    var middleRect: UIView!
+    var bottomRect: UIView!
+
+    var topSpacingRect: UIView!
+    var bottomSpacingRect: UIView!
+
+    let inactiveColor: UIColor
+    let activeColor: UIColor
+    var filterState: ExtendedFilterButtonState
```
whoiswillma: Private fields if possible

Eatery/Views/Eateries/ScrollableTextTabBarControl.swift
```diff
@@ -0,0 +1,104 @@
...
+        select(at: 0)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func select(at index: Int) {
+        // Base case to protect against accidental backend errors
+        if tabButtons.count == 0 {
```
whoiswillma: Slight nit but I think this is the more swifty way to do it: makes clear that you expect `tabButtons` to be non-empty

```suggestion
        guard !tabButtons.isEmpty else {
```

whoiswillma (commented):
😀 lots of amazing work here. Your code is very tidy and well organized. Just a few stylistic / nit-pick comments and some questions. Since this PR is decently large I expect we might go through one or two more rounds of review

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
spd75: To avoid messiness, I didn't want to place the scrollView or the scrollOffset in the constructor.  However, ScrollableViewController still on the scrollOffset and scrollView of its parent view controllers, so within `CampusEateriesViewController`, I actually access the variables publicly and set them in there.  That's why I didn't make those private.  TabBarPadding and filterSpaceHeight can be made private though

spd75: Would you rather me create a setter method in ScrollableViewController that would set the scrollView and scrollOffset and allow them to 1. not be in the constructor to avoid messiness and 2. be private variables?

Or would you prefer some other method of going about this?

Eatery/Utils/Sort.swift
spd75: I assumed implementing merge sort would be more efficient than Swift's sort method.  However, I didn't do much research on the sort function when I did this.  Do you think using swift's sort method would be better?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
spd75: The zPosition property allows for the `HeaderContainer` to be on top of the `ScrollableTextTabBar` when the user decides to sort all items.  Otherwise, the `ScrollableTextTabBar` would be on top when the user sorts everything.

I believe I could've also fixed this by adding the `HeaderContainer` to its superview after adding the `ScrollableTextTabBar`.  However, I decided to do it like this because it would make more sense to add these view in the order they appear on the phone screen.  I could change this though if you'd like.

Eatery Watch App Extension/Queries/campusEateries.graphql
```diff
@@ -49,6 +49,19 @@ query CampusEateries {
                 }
             }
         }
+        expandedMenu {
+        category,
```
chalo2000: remove the comma for consistency? haven't worked with graphql so unsure if its necessary

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
chalo2000: to avoid confusion with what goes on top of what if zPosition happens to be used in the future for the same reason, I would recommend using https://developer.apple.com/documentation/uikit/uiview/1622541-bringsubviewtofront

Eatery/Views/Eateries/ExtendedFilterButton.swift
spd75: Filter state needs to be accessed in another class.  Should I leave it public or add a getter method?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+            make.leading.equalToSuperview().offset(tabBarPadding)
+            make.centerY.equalToSuperview()
+            make.bottom.equalToSuperview()
+        }
+
+        filterButton.snp.makeConstraints { make in
+            make.centerY.equalTo(titleLabel.snp.centerY)
+            make.height.equalTo(titleLabel.snp.height).multipliedBy(0.3)
+            make.trailing.equalToSuperview().offset(-tabBarPadding)
+            make.width.equalTo(filterButton.snp.height).multipliedBy(1.71)
```
chalo2000: 0.3 and 1.71 are very specific, how'd you derive these?

also as an aside, it's nice to put all the constraints as constants somewhere in the file where devs can find it easily instead of scrolling to find where exactly the constraints were made - however, if this is how eatery has been doing constraint constants then just leave it as it is for consistency

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        let titles = viewControllers.map {
+            $0.category
+        }
+        tabBar = ScrollableTextTabBarControl(sections: titles, padding: 16)
+        tabBar.directionalLayoutMargins = NSDirectionalEdgeInsets(top: 0, leading: 16, bottom: 0, trailing: 16)
+        tabBar.addTarget(self, action: #selector(manualScroll), for: .valueChanged)
+        tabBar.backgroundColor = .white
+        tabBar.layer.zPosition = 2
```
chalo2000: and remember to touch on this one too if you end up using bringSubviewToFront

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+        let viewControllerPosOnParent = self.view.frame.origin.y
+        let viewPos = originalViews[tabBar.selectedSegmentIndex].frame.minY
+        let minOffset: CGFloat = 4
+        let targetPos = viewPos - self.scrollOffset + viewControllerPosOnParent + minOffset
+        let topOfSelf = viewControllerPosOnParent - self.scrollOffset
+        let pixPS: CGFloat = 2500
+
+        self.isManualScrolling = true
+
+        if self.scrollView?.contentOffset.y ?? 0 > topOfSelf {
```
chalo2000: you already stored this value in currentPos, use that constant instead

also, instead of using the nil coalescing operator (??) everytime you access currentPos, just use it once when setting currentPos equal to the scrollView's vertical content offset.

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+            let duration = animationDuration(from: currentPos ?? 0, to: targetPos, at: pixPS)
```
chalo2000: this duration is the same as the one in the else block, bring it up to the list of constants so you only write it once

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
spd75: Based off what I've seen from other Eatery files, it seems constraints are hardcoded like they are here, rather than using variables.

I derived those by simple trial and error to try and make the button look as similar to the Figma file as possible.  However, I believe 1.71 was a typo and should be 1.7

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+                self.scrollView?.contentOffset.y = targetPos
+            }
+        } else {
+            let duration = animationDuration(from: currentPos ?? 0, to: targetPos, at: pixPS)
+            let firstProp = fabs((currentPos ?? 0) - topOfSelf) / fabs((currentPos ?? 0) - targetPos)
+            let secondProp = fabs(topOfSelf - targetPos) / fabs((currentPos ?? 0) - targetPos)
+            let firstDuration = duration * Double(firstProp)
+            let secondDuration = duration * Double(secondProp)
+
+            UIView.animate(withDuration: firstDuration, delay: 0, options: .curveEaseIn, animations: {
```
chalo2000: I think some comments explaining what happens in the if and else block would be helpful given there is a lot of math and animations

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+                    }
+                }
+            })
+
+        }
+
+        self.isManualScrolling = false
+    }
+
+    // Function helps animations look more proprotional and smooth
```
chalo2000: proportional

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
whoiswillma: If there's a reason to let them be `internal` then that's good enough for me

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+            make.edges.equalToSuperview()
+        }
+    }
+
+    @objc func manualScroll() {
+        let currentPos = self.scrollView?.contentOffset.y
+        let viewControllerPosOnParent = self.view.frame.origin.y
+        let viewPos = originalViews[tabBar.selectedSegmentIndex].frame.minY
+        let minOffset: CGFloat = 4
+        let targetPos = viewPos - self.scrollOffset + viewControllerPosOnParent + minOffset
```
chalo2000: sometimes you use self and sometimes you don't for the class variables, i'd recommend staying consistent with one style

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+        default:
+            return Double(fabs(cur - target) * (1/pps)) * 1.5
+        }
+    }
+
+    func scrollMenuBar() {
+        let tabBarStackOriginalPos = self.view.frame.origin.y
+        let scrollPos = scrollView?.contentOffset.y ?? 0
+
+        if (scrollPos + scrollOffset) > tabBarStackOriginalPos {
```
chalo2000: this can be factored into a constant to calculate the math once

Eatery/Utils/Sort.swift
whoiswillma: I don't have any references to back me up, but I think Swift's implementation is probably faster (and possibly more correct). I think Swift, like most languages, uses quicksort which can be significantly faster than mergesort due to avoiding an `O(n)` memory allocation

Eatery/Views/Eateries/ExtendedFilterButton.swift
whoiswillma: Can you try `private(set)`?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,374 @@
...
+            let adjustedPosTabBar = (self.headerContainer.frame.height / 2) - (self.tabBar.frame.height / 2)
+            let adjustedPosMenuStack = (self.tabBar.frame.height / 2) + (self.menuItemsView.frame.height / 2)
+            self.separatorView.isHidden = false
+            self.tabBar.isUserInteractionEnabled = false
+            self.tabBar.center.y = self.headerContainer.center.y + adjustedPosTabBar
+            self.menuItemsView.center.y = self.tabBar.center.y + adjustedPosMenuStack
+        }
+
+        if filterButton.filterState == .hightolow {
+            self.scrollView?.contentSize.height -= self.tabBar.frame.height // Does not animate when subtracted
```
chalo2000: is not animating when subtracting a bug or a feature? sounds like a bug TODO

Eatery/Controllers/Eateries/CampusMenuViewController.swift
```diff
@@ -39,14 +39,19 @@ class CampusMenuViewController: EateriesMenuViewController {
...
     private var orderButtonIsHidden = false
     private var scroll: ScrollSession?
     private var showOrderButtonTimer: Timer?
 
+    private var infoView: CampusMenuInfoView!
+    private var scrollableVC: ScrollableViewController?
+
     init(eatery: CampusEatery, userLocation: CLLocation?) {
         self.eatery = eatery
+        self.isExtended = eatery.eateryType == .dining ? false : true
```
chalo2000: `self.isExtended = eatery.eateryType != .dining`

Eatery/Controllers/Eateries/ImageParallaxScrollViewController.swift
```diff
@@ -104,6 +106,15 @@ class ImageParallaxScrollViewController: UIViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
+        // These variables are placed here rather than in CampusMenuVC because
+        // when testing, navBarHeight gave 148 sometimes for some reason...
+        // I'm not sure why this happened but it was 100% correct when put in here
```
chalo2000: could be that it was getting the navBarHeight of the previous screen since here you're guaranteed the views appeared before running the code - maybe find out what code changes the navigationBar's height and put that in viewWillAppear in this file? would be ideal to move these variables to CampusMenuVC if they're unused in this file to have everything in the appropriate place

Eatery/Utils/Sort.swift
chalo2000: https://developer.apple.com/documentation/swift/array/1688499-sort
https://swiftrocks.com/introsort-timsort-swifts-sorting-algorithm

I don't recall if we updated to Swift 5, in which case Swift would be using introsort which has merge sort's complexity but is unstable. If we did update, Swift 5+ is using Timsort which is stable. Stability also doesn't seem to be a problem in this case since I don't believe we care about the order of similarly priced items. I would usually stick to the sorting method that the programming language you're using resorts to unless you have a very good reason for deviating!

Eatery/Views/Eateries/TabBar.swift
```diff
@@ -0,0 +1,42 @@
...
+    init(sections: [String]) {
+        super.init(frame: .zero)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc func tabButtonPressed(_ sender: UIButton) {
+        guard let index = tabButtons.firstIndex(of: sender) else {
```
chalo2000: I'd keep a guard let on the same line by writing the else as `else { return }`

Shared/Models/CampusEatery.swift
```diff
@@ -205,6 +211,14 @@ struct CampusEatery: Eatery, Codable, DefaultsSerializable {
             self.diningMenu = nil
         }
 
+        if let expandMenu = expandedMenu {
+            self.expandedMenu = ExpandedMenu(data: expandMenu)
+        } else {
+            self.expandedMenu = nil
```
chalo2000, line 217: shouldn't expandedMenu already be nil if it skipped the if let?

Shared/Models/CampusEatery.swift
```diff
@@ -255,6 +269,22 @@ extension CampusEatery {
 
 }
 
+// MARK: - ExpandedMenu Information
+
+extension CampusEatery {
+
+    func getOrderedCategories() -> [String] {
+        if let expandedMenu = expandedMenu {
+            let categories = expandedMenu.data.map {
```
chalo2000: `let categories = expandedMenu.data.map { $0.key }`

Shared/Network/NetworkManager.swift
```diff
@@ -156,6 +156,32 @@ struct NetworkManager {
...
+                    expandedMenuData[expandedMenu.category] = []
+                    orderedCategories.append(expandedMenu.category)
+                    var expandedMenuItems: [ExpandedMenu.Item] = []
+                    let stations = expandedMenu.stations.compactMap { $0 }
+
+                    stations.forEach { station in
+                        let items = station.items.compactMap { $0 }
+
+                        items.forEach { item in
+                            let name = item.item
```
chalo2000, line 172: `item.item` is sus, would be good to actually change it to `.name` in a future PR

chalo2000 (approved):
whoa this is big, really good job - i left a couple nitpicks and code improvements and i'll approve so you aren't roadblocked after reviewing my comments 🥇

Shared/Models/CampusEatery.swift
spd75: Since this is in the initializer of a struct, it actually requires that I include `self.expandedMenu = nil` or else it gives a syntax error saying `Return from initializer without initializing all stored properties`

Shared/Network/NetworkManager.swift
spd75: I believe this is actually like that because of the graphQL query needed to get the data from the backend.  So, I think this is a backend issue.

To avoid item.item, though I could name it something like 

`let itemContainers = station.items.compactMap { $0 }`

`itemContainers.forEach { itemContainer in`
       `    let name = itemContainer.item`

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
spd75: Would it be possible for me to create an ExpandedMenus folder within Eatery/Views/Eateries/ ? There are a lot of views in this folder that have to do with Expanded Menus and they're scattered all over the place.  I'm not sure if we could categorize the other views too but I think it would be a solid idea because there's A LOT in this folder.  Let me know what you think

Screen Shot 2021-01-27 at 7 45 17 PM

Eatery Watch App Extension/Queries/campusEateries.graphql
```diff
@@ -49,6 +49,19 @@ query CampusEateries {
...
+            items {
+                item
+                healthy
+                price
+                choices {
+                    options
+                }
+            }
+        }
+        }
```
whoiswillma: Can the block inside `expandedMenu` be indented one level

Eatery/Controllers/Eateries/Campus/CampusEateryExpandedMenuViewControllerswift.swift
```diff
@@ -0,0 +1,50 @@
...
+import UIKit
+
+class CampusEateryExpandedMenuViewController: UIViewController {
+
+    let eatery: CampusEatery
+    let category: String
+    let menu: [ExpandedMenu.Item]
+
+    var contentHeight: CGFloat = 0
+    static let heightConst: CGFloat = 44
```
whoiswillma: Can you throw in a quick doc comment describing what these are for?

Eatery/Controllers/Eateries/Campus/CampusEateryExpandedMenuViewControllerswift.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        var itemViews: [UIView] = []
+
+        for i in 0..<menu.count {
+            let menuRow = ExpandedMenuRow(item: menu[i])
```
whoiswillma:
```suggestion
        for item in menu {
            let menuRow = ExpandedMenuRow(item: item)
```

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+//
+
+import UIKit
+
+class ScrollableViewController: UIViewController {
+
+    private let eatery: CampusEatery!
+    private let viewControllers: [CampusEateryExpandedMenuViewController]!  // The VCs for each menu section
+    private let originalViews: [UIView]!    // The views of the VCs for each menu section
+    private let allItems: [ExpandedMenu.Item]!  // All menu items combined into array and sorted
```
whoiswillma: These are good comments, can you move them above the variable declaration and with three slashes, like this?

```
/// This is a doc comment
private let var: Type
```

This way they'll show up when doing autocomplete, and also the comment can be extended without running into the 120 col limit

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+
+    init(eatery: CampusEatery, viewControllers: [CampusEateryExpandedMenuViewController], items: [ExpandedMenu.Item]) {
+        self.eatery = eatery
+        self.viewControllers = viewControllers
+        self.originalViews = viewControllers.map { $0.view }
+        self.allItems = items.sorted { item1, item2 in
+            if item1.getNumericPrice() > item2.getNumericPrice() {
+                return false
+            }
+            return true
```
whoiswillma: I think this can be simplified to `item1.getNumericPrice() <= item2.getNumericPrice()`? From there the [docs](https://developer.apple.com/documentation/swift/array/2296815-sorted) for `sorted(by:)` mention that the closure should be irreflexive, so the final should be  `item1.getNumericPrice() < item2.getNumericPrice()`

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+    @objc func manualScroll() {
+        let minOffset: CGFloat = 4
+        let currentPos = scrollView?.contentOffset.y ?? 0
+        let viewControllerPosOnParent = view.frame.origin.y
+        let viewPos = originalViews[tabBar.selectedSegmentIndex].frame.minY
+        let targetPos = viewPos - scrollOffset + viewControllerPosOnParent + minOffset
+        let topOfSelf = viewControllerPosOnParent - scrollOffset
+        let duration = animationDuration(from: currentPos, to: targetPos, withConst: 0.03)
+
+        isManualScrolling = true
```
whoiswillma: Is `isManualScrolling` still useful? Seems like it's set to true here then set to false again at the end of this function

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+
+        isManualScrolling = false
+    }
+
+    /*  This function helps scroll animations look more proportional.  However, smaller distances should travel
+        slower, proportionally, compared to larger distances.  That's why this function is logarithmic rather
+        than linear.  If it were linear, shorter scrolls would look too short and choppy.   */
+    private func animationDuration(from cur: CGFloat, to target: CGFloat, withConst propConstant: CGFloat) -> Double {
+        let distance = fabs(cur - target)
+        return Double(log(distance) * log(distance) * propConstant) // Prop constant should be less than 0.1
```
whoiswillma: Can you explain why `propConstant` should be less than `0.1`?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    private func getCompiledItems() -> [ExpandedMenu.Item] {
+        var allMenus: [ExpandedMenu.Item] = []
+        for controller in viewControllers {
+            allMenus += controller.menu
+        }
+
+        return allMenus
+    }
```
whoiswillma: Is this used?

Eatery/Controllers/Eateries/CampusMenuViewController.swift
```diff
@@ -39,14 +39,19 @@ class CampusMenuViewController: EateriesMenuViewController {
     }
 
     private let eatery: CampusEatery
+    private let isExtended: Bool    // Property that tracks whether an ExpandedMenu will be used
```
whoiswillma: Same thing about doc comments here

Eatery/Controllers/Eateries/ImageParallaxScrollViewController.swift
```diff
@@ -17,7 +17,8 @@ class ImageParallaxScrollViewController: UIViewController {
     private let navigationBar = UINavigationBar()
     private let navigationBarBackground = UIView()
 
-    private let scrollView = UIScrollView()
+    internal let scrollView = UIScrollView()
+    internal var scrollOffset: CGFloat?
```
whoiswillma: Variables are `internal` by default
```suggestion
    let scrollView = UIScrollView()
    var scrollOffset: CGFloat?
```

Eatery/Views/Eateries/EateryScreenViews/ExpandedMenus/ExpandedHeaderContainer.swift
```diff
@@ -0,0 +1,92 @@
...
+        separatorView.snp.makeConstraints { make in
+            make.leading.trailing.bottom.equalToSuperview()
+        }
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // Describe why you did this here
```
whoiswillma:
```suggestion
```

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,283 @@
...
+
+    private func switchToHighToLowStack() {
+        if !orderedViewsCreated {
+            menuItemsView.createHighToLowStack()
+        }
```
whoiswillma: Is this supposed to set `orderedViewsCreated = true`? 

I think the more idiomatic way would be to make a method `createOrderedViewsIfNeeded()` and abstract this `if` statement and the one below

Eatery/Controllers/Eateries/Campus/CampusEateryExpandedMenuViewControllerswift.swift
```diff
@@ -0,0 +1,50 @@
...
+//  CampusEateryExtendedMenuViewController.swift
+//  Eatery
+//
+//  Created by Sergio Diaz on 12/20/20.
+//  Copyright © 2020 Cornell AppDev. All rights reserved.
+//
+
+import UIKit
+
+class CampusEateryExpandedMenuViewController: UIViewController {
```
whoiswillma: Should this be converted into a view? It's usage feels much more like a view to be honest

Eatery/Views/Eateries/EateryScreenViews/MenuTabBars/ScrollableTextTabBarControl.swift
```diff
@@ -0,0 +1,104 @@
...
+//  Created by Sergio Diaz on 11/11/20.
+//  Copyright © 2020 Cornell AppDev. All rights reserved.
+//
+
+import UIKit
+import SnapKit
+
+class ScrollableTextTabBarControl: TabBar {
+
+    let padding: CGFloat!
```
whoiswillma:
```suggestion
    private let padding: CGFloat
```

whoiswillma (commented):
Getting there, a few more changes to make and comments to address

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
spd75: Because distances a scroll will travel are usually about 100-3000 pixels, ln(3000)^2 = 32.5 (which is actually log() in swift), this would be WAYY too long for an animation.  Setting prop constant less than 0.1, would make the animation a more reasonable number (3.25 at the most).

spd75: Do you want me to explain this in my code or were you just asking on github?

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+
+import UIKit
+
+class ScrollableViewController: UIViewController {
+
+    private let eatery: CampusEatery!
+    /// The views for each menu section
+    private let categoryViews: [ExpandedMenuCategoryView]!
+    // All menu items combined into array and sorted
+    private let allItems: [ExpandedMenu.Item]!
```
whoiswillma:
```suggestion
    private let eatery: CampusEatery
    /// The views for each menu section
    private let categoryViews: [ExpandedMenuCategoryView]
    // All menu items combined into array and sorted
    private let allItems: [ExpandedMenu.Item]
```

Eatery/Controllers/Eateries/Campus/ScrollableViewController.swift
```diff
@@ -0,0 +1,263 @@
...
+            }
+        } else {
+            // firstProp represents the proportion of animation time spent scrolling from currentPos to the top of self
+            // secondProp represents the proportion of animation time spent scrolling from the top of self to targetPos
+            let firstProp = fabs((currentPos) - topOfSelf) / fabs((currentPos) - targetPos)
+            let secondProp = fabs(topOfSelf - targetPos) / fabs((currentPos) - targetPos)
+            let firstDuration = duration * Double(firstProp)
+            let secondDuration = duration * Double(secondProp)
+
+            UIView.animate(withDuration: firstDuration, delay: 0, options: .curveEaseIn, animations: {
```
whoiswillma: To fix lint warning. I think you'll also need to indent the animation block
```suggestion
            UIView.animate(
                withDuration: firstDuration, 
                delay: 0, 
                options: .curveEaseIn, 
                animations: {
```

whoiswillma (approved):
LGTM! Just two more small things before this PR is ready to :shipit: 

For a future PR: can you add a small separator between sections to make it more clear when we transition from e.g. Bagels to Baked Goods?

spd75:
> LGTM! Just two more small things before this PR is ready to :shipit:
> 
> For a future PR: can you add a small separator between sections to make it more clear when we transition from e.g. Bagels to Baked Goods?

For sure.  Stylistically I'm not sure how we would make this look, so I may ask Zain to come up with something.

PR #319, opened 2022-01-17
brb: Disable universal links in BRBConnectionHandler

whoiswillma:
- Disable universal links in the webview associated with the BRBConnectionHandler by using [this hack?](https://stackoverflow.com/questions/38450586/prevent-universal-links-from-opening-in-wkwebview-uiwebview?rq=1)
  - Xcode did a bunch of cleanup as well I guess

### DIFF
--- a/Eatery.xcodeproj/project.pbxproj
+++ b/Eatery.xcodeproj/project.pbxproj
@@ -1635,7 +1635,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 2.11;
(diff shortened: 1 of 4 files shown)


spd75 (commented):
Did you see the swiftlint comments towards the bottom? There are quite a few. Everything looks good except for those warnings, not sure what you want to do about those.

whoiswillma:
None of them are in files that I touched so I don't particularly want to fix them. 

They ought to be fixed in another PR, in my opinion

spd75 (approved):
Makes sense. Everything else looks good!

